PR #63, opened 2020-02-24
DEV-6922: fix threading issue in LusidApiFactory

chrisbrook83:
# Pull Request Checklist

- [x] Read the [contributing guidelines](https://github.com/finbourne/lusid-sdk-python-preview/blob/master/CONTRIBUTING.md)
- [x] Tests pass

# Description of the PR

Fixes threading issue in `LusidApiFactory` when used in an async context. The previous code assumed `ManagedThreadId` would be constant during an individual (single-threaded) request, but this is not the case if an async continuation (e.g. following an `await`) is employed in the calling application

### DIFF
--- a/sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
+++ b/sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
@@ -20,7 +20,11 @@ namespace Lusid.Sdk.Utilities
     /// <inheritdoc />
     public class LusidApiFactory : ILusidApiFactory
     {
-        private Dictionary<Type, IApiAccessor> _apis;
+        private static readonly IEnumerable<Type> ApiTypes = Assembly.GetAssembly(typeof(ApiClient))
+            .GetTypes()
+            .Where(t => typeof(IApiAccessor).IsAssignableFrom(t) && t.IsClass);
+
+        private readonly IReadOnlyDictionary<Type, IApiAccessor> _apis;
 
         /// <summary>
         /// Create a new factory using the specified configuration
@@ -49,7 +53,7 @@ namespace Lusid.Sdk.Utilities
             
             configuration.AddDefaultHeader("X-LUSID-Application", apiConfiguration.ApplicationName);
 
-            Init(configuration);
+            _apis = Init(configuration);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
```diff
@@ -20,7 +20,11 @@ public interface ILusidApiFactory
     /// <inheritdoc />
     public class LusidApiFactory : ILusidApiFactory
     {
-        private Dictionary<Type, IApiAccessor> _apis;
+        private static readonly IEnumerable<Type> ApiTypes = Assembly.GetAssembly(typeof(ApiClient))
+            .GetTypes()
+            .Where(t => typeof(IApiAccessor).IsAssignableFrom(t) && t.IsClass);
```
chrisbrook83: Made this `static` so it only gets called once per process

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
```diff
@@ -59,17 +63,13 @@ public LusidApiFactory(Configuration configuration)
...
 
-        private void Init(Configuration configuration)
-        {   
-            IEnumerable<Type> apis = Assembly.GetAssembly(typeof(ApiClient))
-                .GetTypes()
-                .Where(t => typeof(IApiAccessor).IsAssignableFrom(t) && t.IsClass);
-
-            _apis = new Dictionary<Type, IApiAccessor>();
-            foreach (var api in apis)
+        private static Dictionary<Type, IApiAccessor> Init(Configuration configuration)
```
chrisbrook83: This change is just so that the `_apis` field can be made readonly

chrisbrook83: This change is just so that the _apis field can be made readonly

sdk/Lusid.Sdk/Utilities/LusidApiFactoryBuilder.cs
```diff
@@ -20,31 +14,20 @@ public static ILusidApiFactory Build(string apiSecretsFilename)
...
-
-                    factory = new LusidApiFactory(config);
-                    ThreadFactories[threadId] = factory;
-                }
+                BasePath = url
+            };
 
-                return factory;
-            }
+            return new LusidApiFactory(config);
```
chrisbrook83: No need to cache these, because:
(a) you DO want a new instance each time, because the factory isn't thread safe, because the contained API instances are not thread safe
(b) these objects are (now) inexpensive to create

PR #68, opened 2020-03-13
DEV-7152: add finaliser to ApiClient to avoid socket leak due to .NET…

chrisbrook83:
Due to a bug in .NET Core 2.2, any HttpClient instances which are not explicitly disposed will leak sockets when running on Linux.

This change ensures the `HttpClient` instances held by `RestClient` instances are always disposed, through use of a finalizer.

### DIFF
--- a/sdk/Lusid.Sdk/Client/ApiClient.Dispose.cs
+++ b/sdk/Lusid.Sdk/Client/ApiClient.Dispose.cs
@@ -1,0 +1,30 @@
+﻿using System;
+
+namespace Lusid.Sdk.Client
+{
+    // DEV-7152: Bug in .NET Core 2.2 means any un-disposed HttpClients will leak sockets
+    // https://github.com/dotnet/corefx/pull/38499
+    // https://github.com/dotnet/runtime/issues/29327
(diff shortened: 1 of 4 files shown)


sdk/Lusid.Sdk/Client/ApiClient.Dispose.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace Lusid.Sdk.Client
+{
+    // DEV-7152: Bug in .NET Core 2.2 means any un-disposed HttpClients will leak sockets
+    // https://github.com/dotnet/corefx/pull/38499
+    // https://github.com/dotnet/runtime/issues/29327
+    // Use a finalizer to ensure RestClient (and thus HttpClient) instances are always disposed
+
+    public partial class ApiClient : IDisposable
```
rizwansaeed: `ApiClient` is one of the auto-generated classes, and is regenerated each time. This implementation will need to be moved to the `utilities` folder which is excluded from being wiped on each run

chrisbrook83: Thanks - move the file

PR #88, opened 2020-07-09
DEV-8692: adding scope and code to flow conv

jz-finbourne:
# Pull Request Checklist

- [ ] Read the [contributing guidelines](../docs/CONTRIBUTING.md)
- [ ] Tests pass

# Description of the PR


sdk/Lusid.Sdk.Tests/tutorials/Instruments/ComplexInstruments.cs
```diff
@@ -50,6 +50,8 @@ public void DemonstrateCreationOfSwap()
 
             // CREATE the flow conventions, index convention
             var flowConventions = new FlowConventions(
+                scope: null,
```
rizwansaeed, line 53: why do you have to explicitly set this to null? it should default to null if not required. what is the default in other languages?

PR #95, opened 2020-08-26
SE-303 LUSID feature reporter

slacki123:
…view

# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Full attribute scanner scans the entire codebase for methods with [LusidFeature("code")] attributes and returns code values as a features.txt file inside the Sdk folder. All tests pass. Please review

### DIFF
--- a/sdk/Lusid.Sdk.Tests/Lusid.Sdk.Tests.csproj
+++ b/sdk/Lusid.Sdk.Tests/Lusid.Sdk.Tests.csproj
@@ -7,6 +7,7 @@
     <WarningsAsErrors />
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="LusidFeatures" Version="0.1.3" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
     <PackageReference Include="Moq" Version="4.12.0" />
     <PackageReference Include="NUnit" Version="3.9.0" />
--- a/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Bitemporal.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Bitemporal.cs
@@ -1,11 +1,11 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using Lusid.Sdk.Api;
 using Lusid.Sdk.Model;
 using Lusid.Sdk.Tests.Utilities;
 using Lusid.Sdk.Utilities;
+using LusidFeatures;
 using NUnit.Framework;
 
 namespace Lusid.Sdk.Tests.Tutorials.Ibor
@@ -28,6 +28,7 @@ namespace Lusid.Sdk.Tests.Tutorials.Ibor
         }
         
         [Test]
+        [LusidFeature("F1")]
         public void Apply_Bitemporal_Portfolio_Change()
         {
             var portfolioId = _testDataUtilities.CreateTransactionPortfolio(TestDataUtilities.TutorialScope);
--- a/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Holdings.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Holdings.cs
@@ -5,6 +5,7 @@ using Lusid.Sdk.Api;
 using Lusid.Sdk.Model;
 using Lusid.Sdk.Tests.Utilities;
 using Lusid.Sdk.Utilities;
+using LusidFeatures;
 using NUnit.Framework;
 
 namespace Lusid.Sdk.Tests.Tutorials.Ibor
@@ -27,7 +28,8 @@ namespace Lusid.Sdk.Tests.Tutorials.Ibor
             var instrumentsLoader = new InstrumentLoader(apiFactory);
             _instrumentIds = instrumentsLoader.LoadInstruments().OrderBy(x => x).ToList();
         }
-
+        
+        [LusidFeature("F2")]
         [Test]
         public void Get_Holdings()
         {
@@ -97,7 +99,8 @@ namespace Lusid.Sdk.Tests.Tutorials.Ibor
             Assert.That(holdings.Values[4].Units, Is.EqualTo(100.0));
             Assert.That(holdings.Values[4].Cost.Amount, Is.EqualTo(10500.0));
         }
-
+        
+        [LusidFeature("F3")]
         [Test]
         public void Set_Target_Holdings()
         {
--- a/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Orders.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Orders.cs
@@ -3,8 +3,8 @@ using System.Collections.Generic;
 using System.Linq;
 using Lusid.Sdk.Api;
 using Lusid.Sdk.Model;
-using Lusid.Sdk.Tests.Utilities;
 using Lusid.Sdk.Utilities;
+using LusidFeatures;
 using NUnit.Framework;
 
 namespace Lusid.Sdk.Tests.Tutorials.Ibor
@@ -29,7 +29,8 @@ namespace Lusid.Sdk.Tests.Tutorials.Ibor
             _instrumentIds = _instrumentLoader.LoadInstruments();
             _ordersApi = _apiFactory.Api<IOrdersApi>();
         }
-
+        
+        [LusidFeature("F4")]
         [Test]
         public void Upsert_Simple_Order()
         {
@@ -79,6 +80,7 @@ namespace Lusid.Sdk.Tests.Tutorials.Ibor
             Assert.That(upsertResult.Values.All(rl => rl.LusidInstrumentId.Equals(_instrumentIds.First())));
         }
         
+        [LusidFeature("F5")]
         [Test]
         public void Upsert_Simple_Order_With_Unknown_Instrument()
         {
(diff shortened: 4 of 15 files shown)


sdk/Lusid.Sdk.Tests/Features/FeatureExtractor.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (att == null) continue;
+
+                classAttributes.Add(att.Code);
+            }
+            return classAttributes;
+        }
+        
+        public static List<String> GetAllMethodAttributesInNamespace(string nameSpace)
+        {
+            Type[] asmTypes = Assembly.GetExecutingAssembly().GetTypes();
```
rizwansaeed, line 33: you can simplify this with the following which will give you a list of all the features

```
var features = Assembly
                .GetAssembly(typeof(FeatureExtractorTests))
                ?.GetTypes()
                .SelectMany(t => t.GetMethods()
                    .Where(m => m.GetCustomAttributes(typeof(LusidFeature), true)?.Length > 0)
                    .Select(m => m.GetCustomAttributes(typeof(LusidFeature)).Cast<LusidFeature>().First().Code)
                );
```

1.`Assembly.GetAssembly(typeof(FeatureExtractorTests))` gets the Assembly for the type `FeatureExtractorTests`
2. Get all the types
3. `SelectMany` will combine the results for each `Type` into a single list
4. For each `Type`, get the methods and filter where it has a `LusidFeature` attribute then select the `First()` LusidFeature` attribute and extract the code
5. What you get back is an `IEnumerable<string>` which is the list of features

slacki123: I love this. I also added a part where I'm only selected from a specified namespace, and I'm allowing the namespace to be case insensitive - the rationale for that is because I've noticed that the 'Properties' features test is under '..tutorials' namespace with lowercase, and others are under '...Tutorials'. Also collecting the resulting stream into a list

`            var featureList = Assembly
                .GetExecutingAssembly()
                .GetTypes()
                .Where(t => t.Namespace != null && t.Namespace.ToLower().StartsWith(nameSpace.ToLower()))
                .SelectMany(t => t.GetMethods()
                    .Where(m => m.GetCustomAttributes(typeof(LusidFeature), true)?.Length > 0)
                    .Select(m => m.GetCustomAttributes(typeof(LusidFeature)).Cast<LusidFeature>().First().Code)
                ).ToList();`

sdk/Lusid.Sdk.Tests/Features/FeatureTests/FeatureExtractorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Lusid.Sdk.Tests.Features.FeatureTests
+{    
+    [TestFixture]
+    public class FeatureExtractorTests
+    {
+        
+        [Test]
+        public void CheckIfValidFeaturesAreReturnedCorrectly()
+        {
+            List<string> expectedAttributes = new List<string> {"F1", "F2", "F3", "F4", "F5", "F6"};
```
rizwansaeed, line 14: C# convention is to use `var` unless the assignment type is no clear/obvious

sdk/Lusid.Sdk.Tests/Features/LusidFeature.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace Lusid.Sdk.Tests.Features
+{
+    public sealed class LusidFeature : Attribute
```
rizwansaeed, line 5: set the `AttributeUsage` to be for methods

sdk/Lusid.Sdk.Tests/Features/LusidFeature.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public LusidFeature(string code)
+        {
+            Code = code;
```
rizwansaeed, line 9: add validation for non-empty strings

sdk/Lusid.Sdk.Tests/Features/FeatureExtractor.cs
rizwansaeed: c# is case sensitive so you don't need to worry about the namespacing vs. folder (vs python which uses folders)

slacki123: Hmm that's strange, because if you look at the c# sdk and go into tutorials -> ibor, and if you compare 'Properties.cs' versus eg. 'Holdings.cs', you will see that the namespace for Properties is **_Lusid.Sdk.Tests.Tutorials.Ibor_** and for Holdings it's **_Lusid.Sdk.Tests.Tutorials.Ibor_**. When I don't make the namespace search case insensitive, I don't get back the features in 'Properties.cs' file.

rizwansaeed: What I meant was you don't need to filter on the namespace, you just need to search the assembly for any class that has the `LusidFeature` attribute

slacki123: I don't think it's desirable to get the features by manually entering class names, as every time a new class would be added, we would need to add that into the test cases. That's why it's better to filter by namespace. 

In addition to that, we don't want to scan through the entire project for all methods with LusidFeature attribute, because it will also include the LusidFeature attributes that we want to be isolated for unit tests only

rizwansaeed: good point bit the tests, you're right  you probably do need to filter them out

sdk/Lusid.Sdk.Tests/Features/LusidFeature.cs
slacki123: Turns out that if I put empty string validation into the LusidFeature attribute class itself, I am then unable to unit test this, because an exception gets thrown whenever the test loads. I think I will keep it in the "ValidateFeatures" method within the "FeatureExtractor" class, which is also a centralised place that contains all feature validation checks. What do you think?

rizwansaeed:
In general use the Rider squiggly lines to help refactor your code (Alt+Enter), there are often useful learnings to be had there too

sdk/Lusid.Sdk.Tests/Features/LusidFeature.cs
rizwansaeed: I think its better keep the validation in `LusidFeature`. That will also mean you don't need your empty string/null tests as it won't be possible to create them in the first instance (the validation will prevent them)

slacki123: Ok, great! I'll remove the tests for that then

sdk/Lusid.Sdk.Tests/Features/FeatureExtractor.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Lusid.Sdk.Tests.Features
+{
+
+    public static class FeatureExtractor
+    {
+        public static List<string> GetAllMethodAttributesInNamespace(string nameSpace)
```
rizwansaeed, line 11: Return as `IEnumerable<string>`

sdk/Lusid.Sdk.Tests/Features/FeatureExtractor.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            var featureList = Assembly
+                .GetExecutingAssembly()
+                .GetTypes()
+                .Where(t => t.Namespace != null && t.Namespace.ToLower().StartsWith(nameSpace.ToLower()))
+                .SelectMany(t => t.GetMethods()
+                    .Where(m => m.GetCustomAttributes(typeof(LusidFeature), true)?.Length > 0)
+                    .Select(m => m.GetCustomAttributes(typeof(LusidFeature)).Cast<LusidFeature>().First().Code)
+                ).ToList();
```
rizwansaeed, line 20: no need to do this here, returning as a List/Array will materialise the enumeration

sdk/Lusid.Sdk.Tests/Features/FeatureExtractor.cs
```diff
@@ -0,0 +1,43 @@
...
+            
+            
+            ValidateFeatures(featureList);
```
rizwansaeed, line 23: you can replace this with

```
            if (featureList.GroupBy(x => x).Count(g => g.Count() > 1) > 0)
            {
                throw new DuplicateFeatureException("LusidFeature annotations with duplicate values have been found. " +
                                                    "Please make sure no LusidFeature annotations duplicates are present.");
                
            }
```

slacki123: Decided to print the duplicate values and display them to the user, so they're easier to find

```
            var duplicatesEnumerable = features
                .GroupBy(g => g)
                .Where(x => x.Count() > 1)
                .Select(k => k.Key);
            var duplicateValues = string.Join(", ", duplicatesEnumerable.ToList());
            if(duplicateValues.Length > 0){
                throw new DuplicateFeatureException($"LusidFeature annotations with duplicate values {duplicateValues} have been found. " +
                                                    "Please make sure no LusidFeature annotations duplicates are present.");
            }
```

sdk/Lusid.Sdk.Tests/Features/LusidFeature.cs
rizwansaeed: was missing `[AttributeUsage(AttributeTargets.Method)]`, I've added now

PR #105, opened 2020-10-27
SE-444: Adding valuations tutorials

slemasne:
# Pull Request Checklist

- [ ] Read the [contributing guidelines](../docs/CONTRIBUTING.md)
- [ ] Tests pass

# Description of the PR

1. Adding one test for valuing a Bond() via the InLine valuations functionality
2. Creating a new Valuations sub directory within the tutorials folder. I think valuations functionality on LUSID has enough depth to warrant its own sub directory as distinct from the Ibor one.


sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
```diff
@@ -0,0 +1,89 @@
...
+                dayCountConvention: FlowConventions.DayCountConventionEnum.Act365,
+                holidayCalendars: new List<string>(),
+                settleDays: 2,
+                resetDays: 2
+            );
+            
+            // Use the GetValuationOfWeightedInstruments to price our Bond
+            // We price the Bond for a date between the Bond's startDate and maturityDate
+
+            ListAggregationResponse bondValuationResult = _apiFactory.Api<IAggregationApi>().GetValuationOfWeightedInstruments(
```
rizwansaeed, line 47: where are the quotes coming from?

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
```diff
@@ -0,0 +1,89 @@
...
+                new InlineValuationRequest(recipeId: new ResourceId(scope: "MyRecipeScope", code: "default"),
```
rizwansaeed, line 48: Avoid using "My" as identifiers

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
```diff
@@ -0,0 +1,89 @@
...
+                            instrumentType: LusidInstrument.InstrumentTypeEnum.Bond
+                        )
+                        )
+                }));
+            
+            // Test that the valuation engine returns a valid result
+            Assert.That(bondValuationResult.Data, Has.Count.EqualTo(1));
+            
+            // Test that the PV calculation is what we expect
+            Assert.That(bondValuationResult.Data[0]["Holding/default/PV"], Is.EqualTo(10336.986301369863));
```
rizwansaeed, line 86: This might be too sensitive if there are the tweaks to the underlying implementation, perhaps change to `Is.EqualTo(...).Within.(0.001).Percent

rizwansaeed (changes requested):

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
slemasne: No quotes needed, right?
The `Bond()` model has all the economic details needed to calculate the PV.

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
MikeMcGarry, line 86: What data is used in the calculation of the present value of the bond? Will this value change over time as the maturity date is hard coded? As interest rates change?

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
rizwansaeed: Where is the (discount) curve coming from?

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
```diff
@@ -0,0 +1,89 @@
...
+            );
+            
+            // Use the GetValuationOfWeightedInstruments to price our Bond
+            // We price the Bond for a date between the Bond's startDate and maturityDate
+
+            ListAggregationResponse bondValuationResult = _apiFactory.Api<IAggregationApi>().GetValuationOfWeightedInstruments(
+                new InlineValuationRequest(recipeId: new ResourceId(scope: "MyRecipeScope", code: "default"),
+                    valuationSchedule: new ValuationSchedule(
+                        effectiveFrom: new DateTimeOrCutLabel("2020-09-01"),
+                        effectiveAt: new DateTimeOrCutLabel("2020-09-01"),
```
jz-finbourne, line 51: effectiveAt = effectiveFrom is redundant, it would default to just that single day.
tenor = "1D" is by default and can be omitted if you wish

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
slemasne: Added that now. See the `UpsertRateCurves` method.

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
slemasne: Done.

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
slemasne: I rounded it to nearest integer. What do you think?

slemasne: I will send you excel with the calculation. The value will change if the discount curve changes, but we have hard coded that into the test.

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
slemasne: Thanks @jz-finbourne 
I have now removed `tenor`
Question - what's the correct valuationSchedule config if I just want to value for one day? Using effectiveAt and effectiveFrom seems to give me the correct results?

jz-finbourne: For valuation of one day, supplying (just) effectiveAt should be sufficient. 
ValuationSchedule(effectiveAt: someDate, effectiveFrom: someDate) is equivalent to just ValuationSchdule(effectiveAt: someDate).  Either works although latter is probably preferred as it is cleaner and internally runs less code (so more efficient and less error-prone).

slemasne: Thanks for quick response @jz-finbourne 
Updated in latest commit

sdk/Lusid.Sdk.Tests/tutorials/Valuations/InlineBondValuation.cs
jz-finbourne: @MikeMcGarry, you can check out this test https://github.com/finbourne/lusid-sdk-csharp-preview/blob/e6770cd55906e0f2f362159b0d1ba7b4f6e9eaf2/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Valuation.cs#L148

where I did something similar to @slemasne, there it values the bond across a week and check that every day (effectiveAt), the PV of the bond varies.

PR #124, opened 2020-12-07
WEAL-1069: Guard against null reference exceptions when calling GetRequestId

Emmetttt:
The current implementation of GetRequestId throws an exception if the `ex.ErrorContent == null`. This merge request adds a null check before proceeding.


rizwansaeed:
can you raise the PR against `develop`

PR #146, opened 2021-04-22
ANA-160: Sdk example on Bond cash flows.

EiriniMavroudi:
# Pull Request Checklist

- [ Y ] Tests pass
- [ Y ] Raised the PR against the `develop` branch

# Description of the PR

Wrote an example of how to compute fixed rate bond cash flows.

### DIFF
--- a/sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
@@ -29,6 +29,48 @@ namespace Lusid.Sdk.Tests.tutorials.Ibor
                 _apiFactory.Api<IQuotesApi>(),
                 _apiFactory.Api<IStructuredMarketDataApi>());
         }
+        
+        [Test]
+        public void ExamplePortfolioCashFlowsForBonds()
+        {
+            // CREATE portfolio
+            var portfolioScope = Guid.NewGuid().ToString();
+            var portfolioId = _testDataUtilities.CreateTransactionPortfolio(portfolioScope);
+
+            // CREATE bond
+            var bond = InstrumentExamples.CreateExampleBond() as Bond;
+        
+            // UPSERT bond to portfolio and populating stores with required market data
+            var effectiveAt = new DateTimeOffset(2020, 2, 23, 0, 0, 0, TimeSpan.Zero);
+            _testDataUtilities.AddInstrumentsTransactionPortfolioAndPopulateRequiredMarketData(
+                portfolioScope, 
+                portfolioId,
+                effectiveAt,
+                effectiveAt,
+                bond);
+            
+            // CALL api to get cashflows at maturity
+            var maturity = bond.MaturityDate.Value;
+            var cashFlowsAtMaturity = _transactionPortfoliosApi.GetPortfolioCashFlows(
+                portfolioScope,
+                portfolioId,
+                maturity.AddMilliseconds(-1),
+                maturity.AddMilliseconds(1));
+            
+            // CHECK correct number of cashflow at maturity
+            var expectedNumber = 2;
+            Assert.That(cashFlowsAtMaturity.Values.Count, Is.EqualTo(expectedNumber));
+            
+            var cashFlows = cashFlowsAtMaturity.Values.Select(cf => cf)
+                .Select(cf => (cf.PaymentDate, cf.Amount, cf.Currency))
+                .ToList();
+            
+            // Check that expected cash flows at maturity are not 0
+            var allCashFlowsPositive = cashFlows.All(cf => cf.Amount > 0);
+            Assert.That(allCashFlowsPositive, Is.True);
+            
+            _portfoliosApi.DeletePortfolio(portfolioScope, portfolioId);
+        }
 
         [TestCase(true)]
         [TestCase(false)]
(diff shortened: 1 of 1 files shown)


sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
```diff
@@ -29,6 +29,48 @@ public void SetUp()
...
+            // CALL api to get cashflows at maturity
+            var maturity = bond.MaturityDate.Value;
+            var cashFlowsAtMaturity = _transactionPortfoliosApi.GetPortfolioCashFlows(
+                portfolioScope,
+                portfolioId,
+                maturity.AddMilliseconds(-1),
+                maturity.AddMilliseconds(1));
+            
+            // CHECK correct number of cashflow at maturity
+            var expectedNumber = 2;
```
rizwansaeed, line 61: I presume this is coupon and redemption notional? if so could we add a comment for the reader

EiriniMavroudi: Yes, it is the last coupon and the principal returned. I will add a comment and update the PR. 

Thank you for spotting it.

sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
```diff
@@ -76,12 +118,56 @@ public void ExamplePortfolioCashFlowsForFxForwards(bool isNdf)
...
+            var currency = currencyAndAmounts.All(t => t.Currency == bond.DomCcy);
+            var amountsPositive = currencyAndAmounts.All(t => t.Amount > 0);
+            Assert.That(currency, Is.True);
+            Assert.That(amountsPositive, Is.True);
+            
+            // Given the cashflow transactions, we create from them transaction requests and upsert them.
+            var upsertCashFlowTransactions = cashFlows.Values;
+            _transactionPortfoliosApi.UpsertTransactions(portfolioScope, portfolioId, CreateCashFlowTransactionRequest(upsertCashFlowTransactions));
+
+            _portfoliosApi.DeletePortfolio(portfolioScope, portfolioId);
```
rizwansaeed, line 168: can you get the transactions back out to show that they (are presumably) cash transactions?

EiriniMavroudi: I updated the test so that it will get the transactions and check that the Transaction type, the currency and the Unit of cash flows is as expected. However, I could not find an explicit 'cash' type in the Transaction object so as to check specifically for that. Do you have an example in mind?

rizwansaeed: assuming you book them as cash instruments, they will have a LUID of CCY_GBP (i think, otherwise it will be GBP)

jz-finbourne: Riz, we can manually change the transactions so it returns with CCY_GBP (say), would you want us to?

Longer story: I think the problem comes back to Transaction vs TransactionRequest. When the client upserts the transactions, they can add in the identifier they so wish... in the current example, we assigned the InstrumentUid to be the same as the instrument that came in so LUID_ABCD1234 rather thank CCY_GBP - we can if you like for this example. Note that this is a manual step and is (currently anyway) not an automatic thing (should it be?).

rizwansaeed: so the issue with creating a new luid is that the cash being generated from the instrument won't be aggregated with the other cash in the portfolio, therefore I think we need to book it using a cash identifier

rizwansaeed: it might be worth adding a property to store the original instrument that generated the flow, or check with @John-from-Finbourne if there is an existing one you can use

jz-finbourne: Okay, we can book it using CCY_GBP as identifier & we can add the instrument LUID to the transaction identifier.

John-from-Finbourne: How are we choosing the transaction type for these transactions?

We can't assume that everybody is using the standard Finbourne codes.
We might need to introduce something to the transaction config so clients can tag the types they want to use for these situations.

Not sure how straightforward this is. It might need a longer discussion.

rizwansaeed: I believe booking the transaction, and choosing the transaction type, is still down the user, what is being shown here is what the flow looks like end to end

jz-finbourne: The transaction type is whatever that gets returned from the GetUpsertableCashFlow. It looks to be the word "buy" is getting returned. 

The question of what the transaction type is, if not mistaken, is more of a question of what is happening below the API rather than this PR (assuming we use what it is returned, see above and L240 onwards). Unless as Riz says, it is manually changed at the user's discretion.

EiriniMavroudi: @rizwansaeed As you requested, the example asserts that the instrument transaction UId returns a cash transaction and the ID is of the form CCY_{bond.CCY}.

rizwansaeed (approved):

PR #154, opened 2021-05-19
[Don't merge] Open API Generator updates

mneedham:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Generated + manual changes for the update to Open API Generator 5.1

There are a bunch of warnings in the generated code like the following:

> /usr/src/Lusid.Sdk/Model/UpsertSingleStructuredDataResponse.cs(142,21): warning CS8073: The result of the expression is always 'true' since a value of type 'DateTimeOffset' is never equal to 'null' of type 'DateTimeOffset?' [/usr/src/Lusid.Sdk/Lusid.Sdk.csproj]

I've disabled those being reported as errors for now, but maybe we can figure out how to get the generated code to not include null checks. I assume they're only unnecessary if those fields are required via a type annotation but I'm not completely sure about that.


sdk/Lusid.Sdk/Lusid.Sdk.csproj
```diff
@@ -1,8 +1,8 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <Version>0.11.3040</Version>
+    <Version>0.11.3028-MARK</Version>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
MikeMcGarry, line 5: Note that  ".NET Framework doesn't support .NET Standard 2.1. For more information, see the announcement of .NET Standard 2.1" - https://docs.microsoft.com/en-us/dotnet/standard/net-standard

According to this document - "We recommend you target the lowest version of .NET Standard possible. So, after you find the highest .NET Standard version you can target, follow these steps..." It then outlines the steps to find the lowest version possible.

As we are not doing anything fancy, let's follow these steps and find the lowest version of netstandard that we can support.

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
```diff
@@ -82,9 +82,10 @@ public string CreateTransactionPortfolio(string scope)
                 },
                 transactionDate: tradeDate,
                 settlementDate: tradeDate,
-                units: units,
-                transactionPrice: new TransactionPrice(price, TransactionPrice.TypeEnum.Price),
-                totalConsideration: new CurrencyAndAmount(price*units, currency),
+                units: units ?? default(decimal),
+                transactionPrice: new TransactionPrice(price ?? default(decimal), TransactionPrice.TypeEnum.Price),
+                totalConsideration: new CurrencyAndAmount((
+                        price ?? default(decimal)) * (units ?? default(decimal)), currency),
```
MikeMcGarry, line 88: Why did you add these defaults in?

sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
```diff
@@ -52,7 +53,7 @@ public void ExamplePortfolioCashFlowsForBonds()
                 bond);
             
             // CALL api to get cashflows at maturity
-            var maturity = bond.MaturityDate.Value;
+            var maturity = bond.MaturityDate;
```
MikeMcGarry, line 56: Why this change?

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
```diff
@@ -37,7 +40,7 @@ public interface ITransactionPortfoliosApi : IApiAccessor
         /// <param name="adjustHoldingRequest">The selected set of holdings to adjust to the provided targets for the              transaction portfolio.</param>
         /// <param name="reconciliationMethods">Optional parameter for specifying a reconciliation method: e.g. FxForward. (optional)</param>
         /// <returns>AdjustHolding</returns>
-        AdjustHolding AdjustHoldings (string scope, string code, DateTimeOrCutLabel effectiveAt, List<AdjustHoldingRequest> adjustHoldingRequest, List<string> reconciliationMethods = null);
+        AdjustHolding AdjustHoldings(string scope, string code, DateTimeOrCutLabel effectiveAt, List<AdjustHoldingRequest> adjustHoldingRequest, List<string> reconciliationMethods = default(List<string>));
```
MikeMcGarry, line 43: What is the implication of the change from a default value of `null` to `default{type}`?

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
```diff
@@ -1,27 +1,30 @@
...
 using Lusid.Sdk.Model;
 
 namespace Lusid.Sdk.Api
 {
+
     /// <summary>
     /// Represents a collection of functions to interact with the API endpoints
     /// </summary>
-    public interface ITransactionPortfoliosApi : IApiAccessor
+    public interface ITransactionPortfoliosApiSync : IApiAccessor
```
MikeMcGarry, line 27: The addition of `Sync` is going to be a breaking change for the SDK. Do we want to keep it or should we remove `Sync` suffix from the template?

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
```diff
@@ -3487,73 +3426,63 @@ public HoldingsAdjustment GetHoldingsAdjustment (string scope, string code, Date
         /// <param name="effectiveAt">The effective datetime or cut label of the holdings adjustment.</param>
         /// <param name="asAt">The asAt datetime at which to retrieve the holdings adjustment. Defaults to the return the latest              version of the holdings adjustment if not specified. (optional)</param>
         /// <returns>ApiResponse of HoldingsAdjustment</returns>
-        public ApiResponse< HoldingsAdjustment > GetHoldingsAdjustmentWithHttpInfo (string scope, string code, DateTimeOrCutLabel effectiveAt, DateTimeOffset? asAt = null)
+        public Lusid.Sdk.Client.ApiResponse<HoldingsAdjustment> GetHoldingsAdjustmentWithHttpInfo(string scope, string code, DateTimeOrCutLabel effectiveAt, DateTimeOffset? asAt = default(DateTimeOffset?))
```
MikeMcGarry, line 3429: I wonder why these are all prefixed with `Lusid.Sdk.Client` now? Doesn't look as good as before.

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
```diff
@@ -3294,73 +3223,74 @@ public VersionedResourceListOfPortfolioHolding GetHoldings (string scope, string
...
             {
-                localVarHeaderParams["Authorization"] = "Bearer " + this.Configuration.AccessToken;
+                localVarRequestOptions.HeaderParameters.Add("Authorization", "Bearer " + this.Configuration.AccessToken);
             }
 
-            //  set the LUSID header
-            localVarHeaderParams["X-LUSID-SDK-Language"] = "C#";
-            localVarHeaderParams["X-LUSID-SDK-Version"] = "0.11.3040";
-
             // make the HTTP request
```
MikeMcGarry, line 3284: What happened to these headers? Note that they are still in the GetHoldingsWithHttpInfoAsync method.

sdk/Lusid.Sdk/Model/WeightedInstrument.cs
```diff
@@ -1,65 +1,68 @@
...
-        
+
         /// <summary>
         /// The quantity of the instrument that is owned.
         /// </summary>
         /// <value>The quantity of the instrument that is owned.</value>
-        [DataMember(Name="quantity", EmitDefaultValue=false)]
-        public decimal? Quantity { get; set; }
+        [DataMember(Name = "quantity", EmitDefaultValue = true)]
+        public decimal Quantity { get; set; }
```
MikeMcGarry, line 53: What is the signifiance of EmitDefaultValue being changed to `true`? - https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.datamemberattribute.emitdefaultvalue?view=netcore-3.1

It looks to me that all required parameters now have this set to true and for all optional parameters it remains set to false. Looking at the options (https://openapi-generator.tech/docs/generators/csharp-netcore/) I think we can only affect the behaviour of the optional parameters so this change should be fine.

sdk/Lusid.Sdk/Lusid.Sdk.csproj
mneedham: oh that's a cool page! I found some other pages but didn't realise that .NET Framework doesn't support .NET Standard 2.1. So then we may as well go back to .NET Standard 2.0

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
mneedham: Should I do `GetValueOrDefault()` instead?

sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
mneedham: There is no `Value` on that property anymore:

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
mneedham: I guess on this line empty query params get added?

```
            if (reconciliationMethods != null)
            {
                localVarRequestOptions.QueryParameters.Add(Lusid.Sdk.Client.ClientUtils.ParameterToMultiMap("multi", "reconciliationMethods", reconciliationMethods));
            }
```

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
mneedham: I can see those headers in a bunch of places in that file:



I only set it once, so I'm not sure why it'd only be missing in one place (if it is)

sdk/Lusid.Sdk/Model/WeightedInstrument.cs
mneedham: I made this change because there is a change in how the JSON serialisation code works. It's in this function:

```
        private bool ShouldWriteProperty(object? memberValue, JsonObjectContract? containerContract, JsonProperty property)
        {
            if (memberValue == null && ResolvedNullValueHandling(containerContract, property) == NullValueHandling.Ignore)
            {
                return false;
            }
            if (HasFlag(property.DefaultValueHandling.GetValueOrDefault(Serializer._defaultValueHandling), DefaultValueHandling.Ignore)
                && MiscellaneousUtils.ValueEquals(memberValue, property.GetResolvedDefaultValue()))
            {
                return false;
            }
            return true;
        }
```

So what happens is that if one of your fields uses the default value for that field (e.g. `rateOrSpread: 0.00m`) then that field won't be included in the request due to it having the default value...unless you set `EmitDefaultValue = true`. 

It only affects integers and decimals.

sdk/Lusid.Sdk/Lusid.Sdk.csproj
MikeMcGarry: Okay, let's make this 2.0 then!

MikeMcGarry: Or can we go even lower to add even more support?

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
MikeMcGarry: Instead, you should make the arguments in `BuildTransactionRequest` non-nullable by removing the `?`

sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
MikeMcGarry: Makes sense, can't be null anymore!

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
MikeMcGarry: Okay, makes sense

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
MikeMcGarry: We need to add these to the sync methods in the template as well.

sdk/Lusid.Sdk/Model/WeightedInstrument.cs
MikeMcGarry: Makes sense, great!

sdk/Lusid.Sdk/Lusid.Sdk.csproj
mneedham: I tried to go lower, but then we can't use RestSharp:

> <PackageReference Include="RestSharp" Version="106.11.7" />

I guess maybe that only became available in .NET Standard 2.0

mneedham: Seems like it's intentionally only for 2.0 and above - https://github.com/restsharp/RestSharp/issues/992

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
mneedham: This should be fixed now.

sdk/Lusid.Sdk/Lusid.Sdk.csproj
MikeMcGarry: Perfecto!

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
MikeMcGarry: Remove units.Value as well and we are good!

sdk/Lusid.Sdk/Api/TransactionPortfoliosApi.cs
MikeMcGarry: Still missing :(

mneedham: I think they should be there now:

PR #161, opened 2021-06-14
Feature/NOJIRA: Fix template links

JeremyBarber:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

The links to the wiki and Readme docs in the PR and Issue Template files dont actually work when clicked from an actual PR or Issue. Try it by clicking the Contribution link above. It wont take you to a valid page.

This PR fixes the links so they will work correctly when clicked from PRs and Issues at the small cost of breaking the links when clicking them as checked out source files or from the GitHub file viewer. I believe these use cases are less important and the trade-off makes sense.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,7 +9,7 @@ assignees: ''
 
 ## Bug Report Checklist
 
-- [ ] I have checked the existing [documentation](../../wiki) for this SDK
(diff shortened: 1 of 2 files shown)


JeremyBarber:
Also, if someone can confirm that raising against `develop` was correct? Because this PR will only affect templates if it is merged into `master` my personal instinct would be to merge it directly into master and I wont be able to tell if it's working or not until it is. Happy to take input.

rizwansaeed:
This will be fixed in the SDK generator project

PR #169, opened 2021-06-30
ApiConfigurationBuilder can be passed IConfigurationSection

cmdkeen:
# Pull Request Checklist

- [X] Read the [contributing guidelines](../blob/master/docs/CONTRIBUTING.md)
- [ ] Tests pass
- [X] Raised the PR against the `develop` branch

# Description of the PR

Adding option for ApiConfigurationBuilder to be passed an IConfigurationSection to allow consuming code to be more flexible with sourcing config from multiple sources.

I originally tried to overload with a .Build(IConfigurationSection) but the .Build(null) to pick up environment variables causes a compiler error as the type of null can't be inferred. I didn't want to propose an API breaking change so just went with a public .BuildFromConfiguration instead.


mneedham:
This looks good to me, but I think we'll let @rizwansaeed have a look at it as well before merging!

rizwansaeed (approved):
Looks good, thanks for the contribution!

PR #193, opened 2021-09-20
invalid tokens now get new tokens, error messages result in 5 retries…

matthew-boardman:
…. New test actually tests this

# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Using Polly, the code now retries getting the token upon a failure. If the refresh token has expired then a new token is requested. The test for this now reflects an expired token. Code has been slightly refactored so that it works and reads better.

### DIFF
--- a/sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
+++ b/sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
@@ -7,6 +7,8 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
+using Polly;
+using Polly.Retry;
 
 [assembly: InternalsVisibleTo("Lusid.Sdk.Tests")]
 
@@ -38,15 +40,17 @@ namespace Lusid.Sdk.Utilities
 
         internal class AuthenticationToken
         {
-            public AuthenticationToken(string token, DateTimeOffset expiresOn, string refreshToken)
+            public AuthenticationToken(string token, DateTimeOffset expiresOn, string refreshToken, DateTimeOffset refreshExpiresOn)
             {
                 Token = token;
                 ExpiresOn = expiresOn;
                 RefreshToken = refreshToken;
+                RefreshExpiresOn = refreshExpiresOn;
             }
-            public string Token { get; }
+            public string Token { get; internal set; }
             public DateTimeOffset ExpiresOn { get; internal set; }
             public string RefreshToken { get; internal set; }
+            public DateTimeOffset RefreshExpiresOn { get; internal set; }
         }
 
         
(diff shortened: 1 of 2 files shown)


sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -63,26 +67,47 @@ public ClientCredentialsFlowTokenProvider(ApiConfiguration configuration)
...
-            await _semaphore.WaitAsync();
-            try
+            AsyncRetryPolicy _policy;
+            _policy = 
+                Policy
+                    .Handle<HttpRequestException>()
+                    .WaitAndRetryAsync(5, retryAttempt =>
+                        TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)), OnRetry);
+            
+            var r = await _policy.ExecuteAsync(context => GetAuthenticationTokenAsyncInternal(), CancellationToken.None);
```
florindr, line 77: you can do without this var r and just return await...

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -63,26 +67,47 @@ public ClientCredentialsFlowTokenProvider(ApiConfiguration configuration)
...
+                    _semaphore.Release();
+                }
+                return _lastIssuedToken.Token;
             }
-            finally
+        }
+
+        private void OnRetry(Exception arg1, TimeSpan arg2)
+        {
+            if (arg1.Message.Contains("refresh token is invalid or expired"))
```
florindr, line 107: add it to a private readonly string class variable and optionally see if there's a section for error messages in the solution to add it there with the rest (to avoid hardcoded strings in the code)

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -125,17 +150,18 @@ private static async Task<AuthenticationToken> GetNewToken(ApiConfiguration apiC
                 // Send request
                 var response = await httpClient.SendAsync(tokenRequest);
                 var body = await response.Content.ReadAsStringAsync();
-
-                if (!response.IsSuccessStatusCode)
+                 
+                if ((int) response.StatusCode != 200)
```
florindr, line 154: leave the .IsSuccessStatusCode to avoid hardcoding that 200

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -125,17 +150,18 @@ private static async Task<AuthenticationToken> GetNewToken(ApiConfiguration apiC
...
                         $"Could not retrieve an authentication token from the specified identity provider. The request to {tokenRequest.RequestUri} returned an unsuccessful status code of {response.StatusCode} and the response body: {body}");
                 }
-
+                
                 var parsed = JsonConvert.DeserializeObject<Dictionary<string, string>>(body);
 
-                var apiToken = parsed["access_token"];                
+                var apiToken = parsed["access_token"];
                 var expires = parsed["expires_in"];
+                var refreshExpires = "5400"; // Refresh token expires in 90 minutes - Speak to Xan if you think this has changed
```
florindr, line 164: promote this to a class variable that's readonly and make it an int so you don't have to do tryparse below

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -201,6 +233,7 @@ private static async Task<AuthenticationToken> RefreshToken(ApiConfiguration api
 
                 var apiToken = parsed["access_token"];                
                 var expires = parsed["expires_in"];
+                var refreshExpires = "5400";
```
florindr, line 236: same as above

florindr (changes requested):

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -63,26 +67,47 @@ public ClientCredentialsFlowTokenProvider(ApiConfiguration configuration)
         /// <inheritdoc />
         public async Task<string> GetAuthenticationTokenAsync()
         {
-            await _semaphore.WaitAsync();
-            try
+            AsyncRetryPolicy _policy;
```
rizwansaeed, line 70: `_` are typically convention for class level variables rather than local method scoped ones

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
rizwansaeed, line 107: it would be more robust if this was based on the http status code, presumably anything other that a 200 range response should be considered a failure

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
```diff
@@ -63,26 +67,47 @@ public ClientCredentialsFlowTokenProvider(ApiConfiguration configuration)
...
             }
-            finally
+        }
+
+        private void OnRetry(Exception arg1, TimeSpan arg2)
+        {
+            if (arg1.Message.Contains("refresh token is invalid or expired"))
             {
-                _semaphore.Release();
+                ExpireRefreshToken();
```
rizwansaeed, line 109: do you not need to expire the token too?

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
matthew-boardman: if we did this for every error then it would have to get a whole new token every time instead of using the refresh token. I don't think this behavior would be wanted for something like a 429?

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
matthew-boardman: I don't think this code would be running of the original token wasn't already expired based on the logic in GetAuthenticationTokenAsyncInternal

sdk/Lusid.Sdk/Utilities/ClientCredentialsFlowTokenProvider.cs
rizwansaeed: if we are doing string comparison, I would also make it lowercase to account for any rogue capitalisation

florindr (approved):

PR #201, opened 2021-10-12
WEAL-2020: fix invalid response from api parsing

slacki123:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

### DIFF
--- a/sdk/Lusid.Sdk/Utilities/ApiExceptionExtensions.cs
+++ b/sdk/Lusid.Sdk/Utilities/ApiExceptionExtensions.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Collections.Generic;
 using System.Net;
 using Lusid.Sdk.Client;
 using Lusid.Sdk.Model;
@@ -33,7 +35,7 @@ namespace Lusid.Sdk.Utilities
             details = null;
             return false;
         }
-        
+
         /// <summary>
         /// Return the details of a validation problem
         /// </summary>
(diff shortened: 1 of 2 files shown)


sdk/Lusid.Sdk/Utilities/ApiExceptionExtensions.cs
```diff
@@ -61,11 +64,26 @@ public static LusidProblemDetails ProblemDetails(this ApiException ex)
...
+                return new LusidProblemDetails(
+                    name: "Invalid JSON error content response",
+                    errorDetails: new List<Dictionary<string, string>>
+                    {
+                        new Dictionary<string, string>
+                        {
+                            {"errorContent", ex.ErrorContent.ToString()}
+                        }
+                    },
+                    detail: $"The error content response is not deserializable: {ex.ErrorContent}");
```
MikeMcGarry, line 85: I think in this case the return should be `null` the same as when there is no error content.

sdk/Lusid.Sdk/Utilities/ApiExceptionExtensions.cs
```diff
@@ -61,11 +64,26 @@ public static LusidProblemDetails ProblemDetails(this ApiException ex)
...
-            });
+                // ApiException.ErrorContent contains a JSON serialized ProblemDetails
+                return JsonConvert.DeserializeObject<LusidProblemDetails>(
+                    ex.ErrorContent.ToString(), new PropertyBasedConverter()
+                );
+            }
+            catch (JsonException)
+            {
+                return new LusidProblemDetails(
+                    name: "Invalid JSON error content response",
```
MikeMcGarry, line 77: I think a user might confuse this with the cause of the ApiException, I think this name should be the name of the `ApiException`. Perhaps the name of the HTTP Status code e.g. `InternalServerError`, `ServiceUnavailable` etc. would be a better fit.

sdk/Lusid.Sdk/Utilities/ApiExceptionExtensions.cs
```diff
@@ -61,11 +64,26 @@ public static LusidProblemDetails ProblemDetails(this ApiException ex)
...
+                    errorDetails: new List<Dictionary<string, string>>
+                    {
+                        new Dictionary<string, string>
+                        {
+                            {"errorContent", ex.ErrorContent.ToString()}
```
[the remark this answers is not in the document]
MikeMcGarry: How does that compare with the keys that are usually found in this dictionary?

PR #217, opened 2021-11-02
ANA-687: Adding a CDS pricing and cash flow test example in Lusid.Sdk.

EiriniMavroudi:
# Pull Request Checklist

- [X] Read the [contributing guidelines](../blob/master/docs/CONTRIBUTING.md)
- [X] Tests pass
- [X] Raised the PR against the `develop` branch

# Inserted an Example of creating a CDS instrument.
Inserted unit tests for checking the CDS pricing and cashflows.

### DIFF
--- a/sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
@@ -91,14 +91,15 @@ namespace Lusid.Sdk.Tests.tutorials.Instruments
                 portfolioScope,
                 discountingRecipeCode);
             
-            // CHECK correct number of cashflow at CDS maturity: There is 1 cash flows corresponding to the last coupon amount.
+            // CHECK correct number of CDS premium leg cashflows at maturity: If no default event is triggered and CDS reaches maturity then there is 1 cash flow expected, which is the last
+            // coupon payment of the protection leg. 
             var expectedNumber = 1;
-            Assert.That(cashFlowsAtMaturity.Values.Count, Is.EqualTo(expectedNumber));
+            Assert.That(cashFlowsAtMaturity.Values.Count, Is.GreaterThanOrEqualTo(expectedNumber));
             
             var cashFlows = cashFlowsAtMaturity.Values.Select(cf => cf)
                 .Select(cf => (cf.PaymentDate, cf.Amount, cf.Currency))
                 .ToList();
-            
+
             // CHECK that expected cash flows at maturity are not 0.
             var allCashFlowsPositive = cashFlows.All(cf => cf.Amount > 0);
             Assert.That(allCashFlowsPositive, Is.True);
--- a/sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
@@ -1,0 +1,134 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Lusid.Sdk.Api;
+using Lusid.Sdk.Model;
+using Lusid.Sdk.Tests.Tutorials.Ibor;
+using Lusid.Sdk.Tests.Utilities;
+using Lusid.Sdk.Utilities;
+using NUnit.Framework;
+
+namespace Lusid.Sdk.Tests.tutorials.Instruments
+{
+    public class CreditDefaultSwapCashFlow
+    {
+        private ILusidApiFactory _apiFactory;
+        private TestDataUtilities _testDataUtilities;
+        private ITransactionPortfoliosApi _transactionPortfoliosApi;
+        private IPortfoliosApi _portfoliosApi;
+        private static IConfigurationRecipeApi _recipeApi;
+        private IAggregationApi _aggregationApi;
+        
+        private static readonly string ValuationDateKey = "Analytic/default/ValuationDate";
+        private static readonly string ValuationPv = "Valuation/PV/Amount";
+        private static readonly string ValuationCcy = "Valuation/PV/Ccy";
+        private static readonly string Luid = "Instrument/default/LusidInstrumentId";
+        private static readonly string InstrumentName = "Instrument/default/Name";
+        private static readonly string InstrumentTag = "Analytic/default/InstrumentTag";
+        
+        private static readonly List<AggregateSpec> ValuationSpec = new List<AggregateSpec>
+        {
+            new AggregateSpec(ValuationDateKey, AggregateSpec.OpEnum.Value),
+            new AggregateSpec(ValuationPv, AggregateSpec.OpEnum.Value),
+            new AggregateSpec(ValuationCcy, AggregateSpec.OpEnum.Value),
+            new AggregateSpec(InstrumentName, AggregateSpec.OpEnum.Value),
+            new AggregateSpec(InstrumentTag, AggregateSpec.OpEnum.Value),
+            new AggregateSpec(Luid, AggregateSpec.OpEnum.Value)
+        };
+
+        [OneTimeSetUp]
+        public void SetUp()
+        {
+            _apiFactory = TestLusidApiFactoryBuilder.CreateApiFactory("secrets.json");
+            _transactionPortfoliosApi = _apiFactory.Api<ITransactionPortfoliosApi>();
+            _portfoliosApi = _apiFactory.Api<IPortfoliosApi>();
+            _recipeApi = _apiFactory.Api<IConfigurationRecipeApi>();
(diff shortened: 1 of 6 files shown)


sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
```diff
@@ -0,0 +1,134 @@
...
+            var expectedNumber = 1;
+            Assert.That(cashFlowsAtMaturity.Values.Count, Is.EqualTo(expectedNumber));
+            
+            var cashFlows = cashFlowsAtMaturity.Values.Select(cf => cf)
+                .Select(cf => (cf.PaymentDate, cf.Amount, cf.Currency))
+                .ToList();
+            
+            // CHECK that expected cash flows at maturity are not 0.
+            var allCashFlowsPositive = cashFlows.All(cf => cf.Amount > 0);
+            Assert.That(allCashFlowsPositive, Is.True);
```
jz-finbourne, line 104: can we remove 
```
            var cashFlows = cashFlowsAtMaturity.Values.Select(cf => cf)
                .Select(cf => (cf.PaymentDate, cf.Amount, cf.Currency))
                .ToList();

            // CHECK that expected cash flows at maturity are not 0.
            var allCashFlowsPositive = cashFlows.All(cf => cf.Amount > 0);
            Assert.That(allCashFlowsPositive, Is.True);
```
this would change and break when protection leg goes in.

Same really with 
`Assert.That(cashFlowsAtMaturity.Values.Count, Is.EqualTo(expectedNumber));`.... maybe
`Assert.That(cashFlowsAtMaturity.Values.Count, Is.GreaterOrEqualTo(1))`;
but tbh, that can break too....

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapPricing.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            // CALL valuation
+            var valuation = _apiFactory.Api<IAggregationApi>().GetValuation(valuationRequest);
+            Assert.That(valuation, Is.Not.Null);
+            Assert.That(valuation.Data.Count, Is.EqualTo(1));
+
+            // CHECK PV - note that swaps/forwards can have negative PV
+            var pv = (double) valuation.Data.First()[HoldingPvKey];
+            Assert.That(pv, Is.Not.Null);
+            Assert.That(pv, Is.GreaterThanOrEqualTo(0));
```
jz-finbourne, line 87: think we don't want this assertion either.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapPricing.cs
```diff
@@ -0,0 +1,90 @@
...
+            // UPSERT the above instrument to portfolio as well as populating stores with required market data
+            _testDataUtilities.AddInstrumentsTransactionPortfolioAndPopulateRequiredMarketData(
+                scope, 
+                portfolioId,
+                TestEffectiveAt,
+                TestEffectiveAt,
+                instrument);
+            
+            // UPSERT CDS spread curve before upserting recipe
+            var cds = (CreditDefaultSwap) instrument;
```
jz-finbourne, line 61: cast this upfront on L50?
i.e. 
`var cds = (CreditDefaultSwap) IE.CreateExampleCDS()`;

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapPricing.cs
```diff
@@ -0,0 +1,90 @@
...
+using Lusid.Sdk.Model;
+using Lusid.Sdk.Tests.Tutorials.Ibor;
+using Lusid.Sdk.Tests.Utilities;
+using Lusid.Sdk.Utilities;
+using NUnit.Framework;
+
+namespace Lusid.Sdk.Tests.tutorials.Instruments
+{
+    [TestFixture]
+    public class CDSPricing
```
jz-finbourne, line 14: -> CreditDefaultSwapPricing

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
```diff
@@ -0,0 +1,134 @@
...
+        
+        [Test]
+        public void ExamplePortfolioCashFlowsForCreditDefaultSwaps()
+        {
+            // CREATE portfolio
+            var portfolioScope = Guid.NewGuid().ToString();
+            var portfolioId = _testDataUtilities.CreateTransactionPortfolio(portfolioScope);
+
+            // CREATE CDS
+            var cds = InstrumentExamples.CreateExampleCreditDefaultSwap() as CreditDefaultSwap;
```
cscialino, line 62: Curious here, and might be a C-sharpism I'm missing, but what does the `as CreditDefaultSwap` do here and is it needed? Does it validate the type or?

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
cscialino: @jz-finbourne why would this break when the protection leg gets created, you expect more cash flows to appear?

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
```diff
@@ -0,0 +1,134 @@
...
+                maturity.AddMilliseconds(-1),
+                maturity.AddMilliseconds(1),
+                null,
+                null,
+                portfolioScope,
+                discountingRecipeCode);
+            
+            // CHECK correct number of cashflow at CDS maturity: There is 1 cash flows corresponding to the last coupon amount.
+            var expectedNumber = 1;
+            Assert.That(cashFlowsAtMaturity.Values.Count, Is.EqualTo(expectedNumber));
```
cscialino, line 96: Can we add something to guide users what cash flows we are returning here? The example looks good, but it's not clear what this 1 cash flow represents. Perhaps even a bit more description on the comment there could help.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
jz-finbourne: It casts the object into the type `CDS`. The CDS's details are used later when upserting market data rather than hardcoding "SNR" for seniority etc

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
jz-finbourne: I would assume/think so. Well, it is certainly possible. 
We don't want to check nor break that here. We test protection leg inside lusid.

cscialino: Do the cash flows returned carry some sort of a type? If we can look say for type "coupon" or "principal", perhaps we can test specifically for that cash flow here. 

And then if the protection leg gets added you'd be able to extend this.

EiriniMavroudi: The test has now been updated and the cash flow is filtered as a premium leg payment.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
EiriniMavroudi: Added further comment for clarification.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapPricing.cs
EiriniMavroudi: Done.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapPricing.cs
EiriniMavroudi: Not necessary. Prefer to keep it as it is.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapPricing.cs
EiriniMavroudi: Removed the second one, kept the one checking that PV is not null.

fightingsleep (approved):

PR #219, opened 2021-11-09
ANA-729: Specify QuoteInterval in rule for ComplexMarketData in preparation for switching on QuoteInterval Filtering

SamuelColvinFinbourne:
…mplexMarketDataAccessor.

# Pull Request Checklist

- [x] Tests pass

# Description of the PR

ANA-729 will turn on functionality to filter ComplexMarketData based on the quoteInterval provided in MarketDataKeyRules given in the ConfigurationRecipe (see [this MR](https://gitlab.finbourne.com/finbourne/lusid/lusid-ibor/-/merge_requests/9054). Currently, quoteInterval does nothing in the complexMarketDataAccessor in Lusid, so we can safely change it on rules used for ComplexMarketData without making a functional change. This allows us to preemtively fix the tests that will break by turning on the filtering. This MR fixes the tests that will break in the C# SDK.

### DIFF
--- a/sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
@@ -113,14 +113,34 @@ namespace Lusid.Sdk.Tests.tutorials.Instruments
         internal static void CreateAndUpsertRecipe(string code, string scope, ModelSelection.ModelEnum model)
         {
             // CREATE a rule for reset quotes
-            var resetRule = new MarketDataKeyRule("Equity.RIC.*", "Lusid", scope, MarketDataKeyRule.QuoteTypeEnum.Price, "mid", quoteInterval: "1Y");
+            var resetRule = new MarketDataKeyRule(
+                key: "Equity.RIC.*",
+                supplier: "Lusid",
+                scope,
+                MarketDataKeyRule.QuoteTypeEnum.Price,
+                field: "mid",
+                quoteInterval: "1Y");
+            var ratesRule = new MarketDataKeyRule(
+                key: "Rates.*.*",
+                supplier: "Lusid",
+                scope,
+                MarketDataKeyRule.QuoteTypeEnum.Price,
+                field: "mid",
+                quoteInterval: "6Y");
+            var creditRule = new MarketDataKeyRule(
+                key: "Credit.*.*.*.*",
+                supplier: "Lusid",
+                scope,
+                MarketDataKeyRule.QuoteTypeEnum.Spread,
+                field: "mid",
+                quoteInterval: "10Y");
             
             // CREATE recipe for pricing
             var pricingOptions = new PricingOptions(new ModelSelection(ModelSelection.LibraryEnum.Lusid, model));
             var recipe = new ConfigurationRecipe(
                 scope,
                 code,
-                market: new MarketContext(new List<MarketDataKeyRule>{resetRule}, options: new MarketOptions(defaultScope: scope)),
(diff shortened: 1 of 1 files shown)


sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
```diff
@@ -113,14 +113,34 @@ public void ExamplePortfolioCashFlowsForCreditDefaultSwaps()
...
+                MarketDataKeyRule.QuoteTypeEnum.Price,
+                field: "mid",
+                quoteInterval: "1Y");
+            var ratesRule = new MarketDataKeyRule(
+                key: "Rates.*.*",
+                supplier: "Lusid",
+                scope,
+                MarketDataKeyRule.QuoteTypeEnum.Price,
+                field: "mid",
+                quoteInterval: "6Y");
```
fbn-seamus, line 129: I'd make this either 1Y or 10Y to match the equity and credit ones. If they can all be 1Y that'd be ideal.

SamuelColvinFinbourne: At least the Rates.*.* needs to be at least 6Y (5Y doesn't work).
Credit.*.*.*.* also works with 6Y.
You just want them uniform? so make them all 6Y? or want them to look round too, so make them all 10Y?

cscialino: Perhaps 10Y works if it needs at least a 5Y lookback. Not sure what mkt data it depends on but might look rounder than 6Y.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
```diff
@@ -113,14 +113,34 @@ public void ExamplePortfolioCashFlowsForCreditDefaultSwaps()
...
+            var creditRule = new MarketDataKeyRule(
+                key: "Credit.*.*.*.*",
```
cscialino, line 131: This is probably fine and how should it work, but out of curiosity, why does it need so many `*` selections?

Looks like previously we had:
- `Equity.RIC.*`
- `Rates.*.*`

Why do we dig so many levels deeper for credit rules?

cscialino: Also, given this isn't exactly clear to me, I think it might be good to add a comment to explain some of the logic behind this which should benefit other end users of the examples as well. 

Maybe a top level that explains the key selection logic/rule sets.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
fbn-seamus: yep, let's go with 10

sdk/Lusid.Sdk.Tests/tutorials/Instruments/CreditDefaultSwapCashFlow.cs
fbn-seamus: The credit mapping allows for a lot more granularity in the depth of the mapping rule, for fairly arbitrary reasons, at least in hindsight. 

I don't think here is the place to go into the nitty gritty on marketdata key rules. But Maybe a comment along the lines of "adding rules with long quote interval windows for cashflow tests, where we are happy to use old market data, as pricing is not a concern"

PR #229, opened 2021-11-23
ANA-686 add ZeroCouponBond example.

zvyagin1:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

I added an example how ZeroCouponBond can be created and used.

### DIFF
--- a/sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
+++ b/sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
@@ -87,6 +87,27 @@ namespace Lusid.Sdk.Tests.Utilities
                 instrumentType: LusidInstrument.InstrumentTypeEnum.Bond
             );
 
+        internal static LusidInstrument CreateExampleZeroCouponBond()
+            => new Bond(
+                startDate: new DateTimeOffset(2020, 2, 7, 0, 0, 0, TimeSpan.Zero),
+                maturityDate: new DateTimeOffset(2020, 9, 18, 0, 0, 0, TimeSpan.Zero),
+                domCcy: "USD",
+                principal: 100m,
+                couponRate: 0m,
+                flowConventions: 
+                    new FlowConventions(
+                        currency: "USD",
+                        paymentFrequency: "0Invalid",
+                        rollConvention: "MF",
+                        dayCountConvention: "Act365",
+                        paymentCalendars: new List<string>(),
+                        resetCalendars: new List<string>(),
+                        settleDays: 2
+                    ),
+                identifiers: new Dictionary<string, string>(),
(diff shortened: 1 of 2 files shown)


sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
```diff
@@ -87,6 +87,27 @@ internal static LusidInstrument CreateExampleBond()
...
+            => new Bond(
+                startDate: new DateTimeOffset(2020, 2, 7, 0, 0, 0, TimeSpan.Zero),
+                maturityDate: new DateTimeOffset(2020, 9, 18, 0, 0, 0, TimeSpan.Zero),
+                domCcy: "USD",
+                principal: 100m,
+                couponRate: 0m,
+                flowConventions: 
+                    new FlowConventions(
+                        currency: "USD",
+                        paymentFrequency: "0Invalid",
```
rizwansaeed, line 100: what does this mean? where does a user get a list of fields that they can use? we need to add either comments for them to resolve or links to a KB article

sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
```diff
@@ -87,6 +87,27 @@ internal static LusidInstrument CreateExampleBond()
                 instrumentType: LusidInstrument.InstrumentTypeEnum.Bond
             );
 
+        internal static LusidInstrument CreateExampleZeroCouponBond()
+            => new Bond(
+                startDate: new DateTimeOffset(2020, 2, 7, 0, 0, 0, TimeSpan.Zero),
+                maturityDate: new DateTimeOffset(2020, 9, 18, 0, 0, 0, TimeSpan.Zero),
+                domCcy: "USD",
+                principal: 100m,
```
rizwansaeed, line 95: what units and cost should transactions/holdings be booked with (its not clear from this example)?

sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
zvyagin1: Unfortunetelly right now this is the only way how a ZCB schedule can be created. `0Invalid` is 

- not documented
- will be replaced in future

sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
zvyagin1: The ZCB creation is more-or-less identical to a bond creation example we already had before. The only _difference_ is that there is a single payment and coupon rate is zero. The asked question (if clarification is requited) should not be answered in the ZCB example, as other instruments use the same  **domCcy**, **principal** parameters.

PR #230, opened 2021-11-25
Feature/seng 54.fix null responses

slacki123:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR


sdk/Lusid.Sdk.Tests/ApiClientTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        public void SetUp()
+        {
+            _apiFactory = TestLusidApiFactoryBuilder.CreateApiFactory("secrets.json");
+            _apiClient = new ApiClient();
+            _customJsonCodec = new CustomJsonCodec(_apiClient.SerializerSettings, GlobalConfiguration.Instance);
+        }
+
+        [Test]
+        [TestCase(" ", "API Response Content was invalid: ' '")]
+        [TestCase("", "API Response Content was invalid: ''")]
```
rizwansaeed, line 32: use String.Empty?

sdk/Lusid.Sdk.Tests/ApiClientTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            var mockRestResponse = new RestResponse
+            {
+                Content = apiResponseString
+            };
+
+            var exception = Assert.Throws(Is.InstanceOf<Exception>(),
+                () => _customJsonCodec.Deserialize(mockRestResponse, typeof(Portfolio))
+            );
+
+            Assert.That(exception.InnerException.Message, Is.Not.EqualTo(""));
```
rizwansaeed, line 46: String.Empty

sdk/Lusid.Sdk.Tests/tutorials/Ibor/Portfolios.cs
```diff
@@ -253,6 +254,7 @@ public void Add_Transactions_To_Portfolio_With_Property()
         
         [LusidFeature("F2-4")]
         [Test]
+        // TODO: These tests names need to be renamed and be in line with the standard https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices
```
rizwansaeed, line 257: these are intended to be working examples rather than tests per se, I dont think they need to follow test naming convention

sdk/Lusid.Sdk.Tests/tutorials/Ibor/Portfolios.cs
```diff
@@ -283,6 +285,16 @@ public void List_Scopes()
 
             Assert.That(scopes.Values.Count(), Is.GreaterThan(0));
         }
+        
+        [Test]
+        public void GetPortfolio_WhenPortfolioNotFound_Throws404Error()
```
rizwansaeed, line 290: I think this should be `_ThrowsApiException`, the test for the 404 behaviour should be in LUSID itself

sdk/Lusid.Sdk/Client/ApiClient.cs
```diff
@@ -3,7 +3,7 @@
...
  *
- * The version of the OpenAPI document: 0.11.3754
+ * The version of the OpenAPI document: 0.11.3757
```
rizwansaeed, line 6: the is a generated file and shouldn't be checked in

sdk/Lusid.Sdk/Client/ApiResponse.cs
```diff
@@ -12,6 +12,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
+using RestSharp;
```
rizwansaeed, line 15: the is a generated file and shouldn't be checked in

sdk/Lusid.Sdk/Client/Configuration.cs
```diff
@@ -51,12 +52,25 @@ public class Configuration : IReadableConfiguration
         public static readonly ExceptionFactory DefaultExceptionFactory = (methodName, response) =>
         {
             var status = (int)response.StatusCode;
+
```
rizwansaeed, line 55: is this a generated class/code?

sdk/Lusid.Sdk.Tests/tutorials/Ibor/Portfolios.cs
slacki123: Yes I agree now, I'll remove the comment

sdk/Lusid.Sdk.Tests/tutorials/Ibor/Portfolios.cs
slacki123: fair point!

sdk/Lusid.Sdk/Client/Configuration.cs
slacki123: Yes this is indeed. We should be able to inject this functionality like Mike suggested in his comment https://finbourne.atlassian.net/browse/SENG-54

sdk/Lusid.Sdk/Client/ApiClient.cs
slacki123: good spot. ApiResponse.cs changes should be made in the generator template though as it is generated

sdk/Lusid.Sdk.Tests/ApiClientTests.cs
slacki123: Inputs into [TestCase(..)]  must be constant expressions, cannot be objects or object method calls, so String.Empty will not compile

sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
```diff
@@ -12,6 +12,7 @@
 using Newtonsoft.Json;
 using NUnit.Framework;
 using System.Net;
+using RestSharp;
```
MikeMcGarry, line 15: Is this necessary?

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            var status = (int)response.StatusCode;
+            
+            // Throw if status code is >= 400
+            if (status >= 400)
+            {
+                return new ApiException(status,
+                    $"Error calling {methodName}: {response.RawContent}",
+                    response.RawContent, response.Headers);
+            }
```
MikeMcGarry, line 28: This already exists on the default exception handler, why not just call that?

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+            if (response.ResponseStatus == ResponseStatus.Error)
+            {
+                return new ApiException(500,
+                    $"Internal SDK error occured when calling {methodName}: {response.ErrorText}",
+                    response.InternalException.StackTrace, response.Headers);
+
+                // TODO: Since we can't log this potentially sensitive information.
+                // consider sending the full response body here into a safe storage for debugging
+            }
+            return null;
```
MikeMcGarry, line 40: It is possible that LUSID responds with a `200` statusCode but we end up with a `500`, do you think this is a good result? Do you think it is misleading?

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
```diff
@@ -75,6 +75,9 @@ public LusidApiFactory(Configuration configuration)
                 {
                     throw new Exception($"Unable to create type {api}");
                 }
+                
+                // Replace the default implementation of the ExceptionFactory with a custom one defined by FINBOURNE
+                impl.ExceptionFactory = LusidExceptionHandler.CustomExceptionFactory;
```
MikeMcGarry, line 80: Nice

sdk/Lusid.Sdk/Client/ApiClient.cs
```diff
@@ -136,15 +136,11 @@ internal object Deserialize(IRestResponse response, Type type)
...
-            }
-            catch (Exception e)
-            {
-                throw new ApiException(500, e.Message);
-            }
+            // If for some reason the response is a null, empty, or whitespace, the serialisation should fail.
+            if (string.IsNullOrWhiteSpace(response.Content))
+                throw new DeserializationException(response, new Exception($"API Response Content was invalid: '{response.Content}'"));
+            
+            return JsonConvert.DeserializeObject(response.Content, type, _serializerSettings);
```
MikeMcGarry, line 143: I'm not convinced that we want to make this change. If we can handle this scenario with a custom exception handler why modify the generated template? Might be better to post an issue on the OpenApi Generator repo instead.

These changes make upgrading the version of OpenApi much more difficult as we need to migrate them over to a potentially modified template and it isn't always clear what their significance is.

sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
slacki123: Needed for the ResponseStatus enum access

sdk/Lusid.Sdk/Client/ApiClient.cs
slacki123: We can't really handle empty or whitespace responses in the exception handler robustly. If we pass an empty string through the Deserializer, we end up with no error thrown and just a 'null' return value, which then becomes a null API method response. So not sure what to do about that in the nearest future.

I do agree raising an issue with Open API

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
slacki123: could do to be fair. But since this is custom, I wanted to keep it separate from any default implementations. What do you think is best?

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
slacki123: Yeah I've thought about it and it's a fair point. Might be better to keep the status code the same as the response on the API. This is still an exception either way

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+            // RawContent is the response body string of the API response.
+            // If an empty string or whitespace, we should throw because Deserialization just returns null on empty strings.
+            if (string.IsNullOrWhiteSpace(responseBodyString))
+                return new ApiException(
+                    status,
+                    $"Internal SDK error occured when calling {methodName}",
+                    $"API response body was invalid: '{responseBodyString}'. " +
+                    $"Please check the logs for this request and potentially raise the issue with the API team.",
+                    headers
+                );
```
MikeMcGarry, line 50: Are you telling me that ex.ErrorContent will be `null` in this case?

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+            var headers = response.Headers; // Http headers that come back with the api response
+            var responseBodyString = response.RawContent; // The body content of the http response as a string
+            
+            // Use default exception, and only use subsequent checks if default exception factory is null.
+            var defaultException = Configuration.DefaultExceptionFactory.Invoke(methodName, response);
+            if (defaultException != null) return defaultException;
+            
+            // Throw whenever an internal SDK error has been thrown.
+            // Internal SDK deserialization errors will result in ErrorText to be not null.
+            if (response.ErrorText != null)
```
MikeMcGarry, line 32: Are we using != null or IsNullOrWhiteSpace? Let's pick one to use for both

sdk/Lusid.Sdk/Client/Configuration.cs
```diff
@@ -513,4 +513,4 @@ public static IReadableConfiguration MergeConfigurations(IReadableConfiguration
         }
         #endregion Static Members
     }
-}
+}
```
MikeMcGarry, line 516: Get this out of the diff just in case it causes us rebase problems as this is a generated file.

sdk/Lusid.Sdk.Tests/LusidExceptionHandlerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        [OneTimeSetUp]
+        public void SetUp()
+        {
+            _apiFactory = TestLusidApiFactoryBuilder.CreateApiFactory("secrets.json");
+        }
+
+
+        [Test]
+        public void CallCustomExceptionFactory_FailsWithInternalError_Returns500Exception()
```
MikeMcGarry, line 24: Change test name to match updated behaviour where statusCode is now 200

sdk/Lusid.Sdk.Tests/LusidExceptionHandlerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            
+            var returnedError = (ApiException) LusidExceptionHandler.CustomExceptionFactory(methodName, response);
+
+            Assert.That(returnedError.Message, Is.EqualTo(expectedErrorContent));
+            Assert.That(returnedError.ErrorCode, Is.EqualTo(404));
+            Assert.That(returnedError.ErrorContent, Is.EqualTo(rawContent));
+        }
+
+        [Test]
+        public void CallPortfoliosApiExceptionFactory_DefaultIsOverriden_ExceptionFactoryOfApiIsSameAsCustomOne()
```
MikeMcGarry, line 64: Nice test

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
slacki123: correct. Passing a "" or " " into the deserializer throws no error...

sdk/Lusid.Sdk.Tests/LusidExceptionHandlerTests.cs
slacki123: thanks!

sdk/Lusid.Sdk/Client/Configuration.cs
slacki123: 👍

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
slacki123: what's the problem with != null?

MikeMcGarry: Perhaps it is helpful to also check if it is just white space, but in this case it sounds like it should always be null if not an error so let's leave it as is.

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
MikeMcGarry: Okay nice catch, cool.

MikeMcGarry:
Approved! Nice addition!

chrisbrook83:
Note, substantial parts of this PR have been included in this change:
https://github.com/finbourne/lusid-sdk-csharp-preview/pull/268

PR #235, opened 2021-12-07
SENG-54: Add default Polly retry policy for internal SDK exceptions

slacki123:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Some points to consider

- There will be up to 2 retries (can easily change the number) on regular API timeouts. (We could potentially not have a retry on regular timeouts by adding a condition check, but not yet sure whether internal SDK exceptions sometimes also result in a timeout that looks like an API timeout)

- We will have stack trace logs for the time being helping us to identify the different types of internal SDK errors and where they originate for information gathering purposes

- Currently the SDK does not handle deserialisation errors. This is also reflected in the tests of the PR above. I will modify them when I raise another PR with the default exception factory to account for those.

- There's a test marked as explicit to show an example of whenever an internal SDK exception might occur. It is explicit because this particular test does not work the same way inside a docker container, but does work when running on my local Windows machine. It is a good example of how a 0 status code can be created though.


sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            _httpListener.Prefixes.Add(ListenerUriPrefix);
+            
+            var testApiConfig = TestLusidApiFactoryBuilder.CreateApiConfiguration(null);
+            testApiConfig.ApiUrl = ListenerUriPrefix;
+            
+            _apiFactory = new LusidApiFactory(testApiConfig);
+        }
+
+        [Test]
+        [Explicit("This passes locally, but not in the pipeline, and no idea why. Setting to explicit until resolved.")]
```
MikeMcGarry, line 34: Try running this test inside a docker container on your local machine, that should give you insight into what this isn't passing in the pipeline but is locally. We can't merge this until resolved and the test is in place.

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+                    MaxRetryAttempts,
+                    onRetry: (result, retryCount, context) =>
+                    {
+                        Console.WriteLine("An internal exception has occurred. Retrying. " +
+                                          $"Retry attempt: {retryCount}. Max retry attempts: {MaxRetryAttempts}");
+
+                        Console.WriteLine(
+                            "Temporarily logging the exception stack trace for finding the underlying root cause of this exception:\n" +
+                            $"{result.Result.ErrorException}\n");
+                    });
```
MikeMcGarry, line 45: Why is this temporary?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
```diff
@@ -0,0 +1,68 @@
...
```
MikeMcGarry, line 66: This looks like code duplication, I suggest delagating this and the one above in the Sync handler to a separate method.

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
```diff
@@ -68,6 +68,15 @@ public LusidApiFactory(Configuration configuration)
 
         private static Dictionary<Type, IApiAccessor> Init(Configuration configuration)
         {
+            // If some retry policy has already been assigned, use it.
+            // Users can combine their own policy with the default policy by using the .Wrap() method.
```
MikeMcGarry, line 72: Can you provide an example of this in your tests?

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
```diff
@@ -68,6 +68,15 @@ public LusidApiFactory(Configuration configuration)
...
+            RetryConfiguration.RetryPolicy = 
+                RetryConfiguration.RetryPolicy ?? ApiRetryHandler.GetInternalExceptionRetryPolicy();
+            
+            // If some async retry policy has already been assigned, use it.
+            RetryConfiguration.AsyncRetryPolicy =
+                RetryConfiguration.AsyncRetryPolicy ?? ApiRetryHandler.GetInternalExceptionAsyncRetryPolicy();
```
MikeMcGarry, line 78: If I instantiate an ApiClientFactory this gets set for every ApiClient both already instantiated and those that will be created in the future, is this correct?

It is it possible to only apply the `ApiRetryHandler` to the instances of the ApiClient produce by this ApiClientFactory?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    /// <summary>
+    /// Class used to define API error retry rules for all API calls.
+    /// </summary>
+    public static class ApiRetryHandler
+    {
+        /// <summary>
+        /// Number of max retry attempts
+        /// </summary>
+        public const int MaxRetryAttempts = 3;
```
MikeMcGarry, line 15: Should this be configurable by a user?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Get the internal exception condition on which to retry.
+        /// </summary>
+        /// <param name="restResponse">Response object that comes from the API Client</param>
+        /// <returns></returns>
+        private static bool GetInternalExceptionRetryCondition(IRestResponse restResponse)
+        {
+            return restResponse.ErrorException != null || restResponse.StatusCode == 0;
```
MikeMcGarry, line 24: Why/when would the statusCode be `0`? In what circumstances does the `restResponse.ErrorException` exist? Will this fire on all non 2xx status codes?

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+                    });
+
+            RetryConfiguration.RetryPolicy = testRetryPolicy;
+
+            // Calling GetPortfolio or any other API triggers the flow that triggers polly
+            var sdkResponse = _apiFactory.Api<IPortfoliosApi>().GetPortfolio("any", "any");
+            
+            Assert.That(currentRetry, Is.EqualTo(expectedNumberOfRetries));
+            // In the future 0 error codes with throw an error after retries exceeded
+            Assert.That(sdkResponse, Is.Null);
```
MikeMcGarry, line 73: So if the maximum retries are reached the response is `null`? Why is that?

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+                 _httpListener.BeginGetContext(result =>
+                {
+                    var listener = (HttpListener) result.AsyncState;
+                    // Call EndGetContext to complete the asynchronous operation.
+                    var context = listener.EndGetContext(result);
+
+                    // Obtain a response object.
+                    var response = context.Response;
+
+                    // Abort the response. This returns 0 status code.
```
MikeMcGarry, line 50: When would this happen in reality? A `0` status code being returned?

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        [Test]
+        public void CreateLusidFactory_WhenRetryPolicyIsNull_AssignsDefaultRetryPolicy()
+        {
+
+            RetryConfiguration.RetryPolicy = null;
+            
+            var newFactory = new LusidApiFactory(new Configuration());
+
+            Assert.That(RetryConfiguration.RetryPolicy, Is.Not.Null);
```
MikeMcGarry, line 84: This is assigning the ApiRetryHandler right? Can you assert on that instead of just that it is not null.

MikeMcGarry: Again, is this the behaviour that we want where we modify the global RetryConfiguration? Or should we just update the config for the ApiClient's produced by the instance of the ApiClientFactory.

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+        
+        [Test]
+        public void CreateLusidFactory_WhenRetryPolicyIsAlreadyAssigned_ExistingRetryPolicyIsUsed()
```
MikeMcGarry, line 88: Okay nice.

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            var testPolicy = Policy.HandleResult<IRestResponse>(response => true).Retry();
+            
+            RetryConfiguration.RetryPolicy = testPolicy;
+            var newFactory = new LusidApiFactory(new Configuration());
+
+            Assert.That(RetryConfiguration.RetryPolicy, Is.EqualTo(testPolicy));
+        }
+
```
MikeMcGarry, line 97: There isn't a single test here that uses the `ApiRetryHandler`. How do you know that it works?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
MikeMcGarry: It's not clear to me at all when a retry will occur and when it won't.

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+            
+            var currentRetry = 0;
+            var testRetryPolicy = Policy
+                .HandleResult<IRestResponse>(restResponse => restResponse.StatusCode == 0)
+                .Retry(
+                    retryCount: expectedNumberOfRetries,
+                    (result, i) =>
+                    {
+                        Console.WriteLine($"If you see this, then polly retry works. Retry number: {i}");
```
MikeMcGarry, line 62: Not suitable for check-in to master. Use assertions instead where you want to check the behaviour.

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: We needs tests which show what happens when:

- There is a 2xx response (it's not clear what happens here currently, prove that we don't retry in this situation)
- There is a non-2xx response (it's not clear what happens here currently, what is your intention with this change?)
- We reach the maximum number of retries (it's not clear what happens here currently, what is your intention with this change?)
- If we can't deserialsie the response  (it's not clear what happens here currently, what is your intention with this change?)

We also need a clear understanding of:

- Whether instantiating an ApiClientFactory affects the global RetryConfiguration and whether or not this is the desired behaviour
- The wrap behaviour that you mentioned and how a client would implement that

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+        public static Policy<IRestResponse> GetInternalExceptionRetryPolicy() =>
+            Policy
+                .Handle<SystemException>()
+                .OrResult<IRestResponse>(GetInternalExceptionRetryCondition)
+                .Retry(
+                    MaxRetryAttempts,
+                    onRetry: (result, retryCount, context) =>
+                    {
+                        Console.WriteLine("An internal exception has occurred. Retrying. " +
+                                          $"Retry attempt: {retryCount}. Max retry attempts: {MaxRetryAttempts}");
```
MikeMcGarry, line 40: This could become quite chatty. Is there a way to implement this using a logging library like you can do in Python where the user defines if they want to display the logs or not?

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: Good shout on using docker. I'll have a go

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: Because when we get these errors, we want to at least know exactly where they originate, so that if we actually find a way to fix it in the future, we know where to start

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: yeah that's fair

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: As in should we allow it to be configurable or is it configurable now?

I don't think users should be able to configure this. I thought a const assignment makes the variable immutable.

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: Status code is 0 whenever for example the http connection is abruptly aborted, or a timeout happens, or any other exception that originates from within the SDK.

When the above happens, an ErrorException object will also be added to the response

It shouldn't trigger on any other status code than 0, unless these staus codes throw an exception internally somewhere

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: because it retries 3 times, and still gets the same response as it got the first time around. Which was null. 

(When we implement the exception factory handler, this will then throw an exception after the 3rd retry)

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: When connection is aborted, or the thing timeouts. Also there are many case scenarios we don't need the root cause of, that's why we want to see the stack trace with these retries, at least temporarily so we are familiar with the failures

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: yeah will have to figure out how to compare policies

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: ApiRetryHandler itself isn't really being tested. I'm also not sure how we could or should test it. If we were testing it, then we'd only be testing the Polly library functionality, but open to suggestions here. What we can test though, is to show that when we assign a Polly retry policy, then it does actually work on our selected errors thrown by the REST client

slacki123: I agree on the test cases you suggested, I'll look into adding those.

I can also add an example how Polly wrap can be used to combine policies I guess.

> Whether instantiating an ApiClientFactory affects the global RetryConfiguration and whether or not this is the desired behaviour
Do you mean LusidApiFactory. Is ApiClientFactory separate? Also the desired behaviour is probably up for discussion

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: fairs

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: This will occur only once in roughly 50k API calls the last time we ran a test in the pipeline. So quite rare. It should work after a single retry.

I'll see if we can wrap these in any logging libraries

Also, we want to make sure that we do have these logs as warnings and an indication that the underlying issue of the SDK still exists. Because retrying is just kind of putting a plaster on it - what do you think?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: What other name would you give to `GetInternalExceptionRetryCondition` to make it more self explanatory?

slacki123: Or maybe it just needs updating the comment?

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
slacki123: I'll try to find a way to make an example test using this

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
slacki123: I was originally looking where we could set a default retry policy on startup. We can't do it in the RetryConfiguration.RetryPolicy, because that's part of the generation template.

I was hoping that by creating any API factory, the retry policy would be assigned.

Would we want it to work differently? And why if yes?

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: > If we can't deserialsie the response (it's not clear what happens here currently, what is your intention with this change?)

Open API doesn't currently support retries on deserialization errors by the looks of it. On those, we can implement ExceptionFactoryHandler and throw if this is the case. 

>

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
slacki123: Done

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: But let me know what you think

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: Added plenty of examples in the latest MR

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
MikeMcGarry: This is a production SDK I don't think we should temporarily write anything to the console. If you want to expose this information we should find a way to allow the user to decide what log level to expose. Here this will be printed whether you want it or not.

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
MikeMcGarry: I can certainly see a use case where you have different ApiClientFactory's with different retry settings. The behaviour might surprise you if you try and change one and they all change.

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: We don't have a logging framework implemented, so will this suffice?

```
        private static void HandleRetryAction(DelegateResult<IRestResponse> result, int retryCount, Context context)
        {
            if (Environment.GetEnvironmentVariable("HIDE_INTERNAL_EXCEPTION_RETRY_LOGS") == "true") return;
            
            Console.WriteLine("An internal exception has occurred. Retrying. " +
                              $"Retry attempt: {retryCount}. Max retry attempts: {MaxRetryAttempts}");

            Console.WriteLine(
                "Temporarily logging the exception stack trace for finding the underlying root cause of this exception:\n" +
                $"{result.Result.ErrorException}\n");
        }
```

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
slacki123: could you give an example of how to remedy this?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: Then they create a polly policy on their own end. and assign/wrap it with the global configuration. These are not for the client to modify as they only handle internal exceptions, such as the 0 status code ones

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
slacki123: Also, if we were to implement configurable logging, it should probably be a separate PR

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
slacki123: `RetryPolicy` config is a global static method used in the generated ApiClient class https://github.com/finbourne/lusid-sdk-csharp-preview/blob/master/sdk/Lusid.Sdk/Client/ApiClient.cs#L492

How would you remedy this such the polly retry policy configuration is different for every LusidApiFactory instance, and where would you set the configuration from?

MikeMcGarry: Fair enough, this behaviour is consistent with the default behaviour of a global retry policy.

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
MikeMcGarry: No that's okay, let's leave it in and we can address the logging in a separate merge request.

sdk/Lusid.Sdk/Utilities/PollyApiRetryHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// </summary>
+        /// <returns>Fallback Policy (Synchronous)</returns>
+        public static Policy<IRestResponse> DefaultFallbackPolicy =>
+            Policy<IRestResponse>
+                .Handle<SystemException>()
+                .Fallback(
+                    (outcome, ctx, ct) => outcome.Result,
+                    (outcome, ctx) =>
+                    {
+                        // Add logging or other logic here
```
MikeMcGarry, line 60: Are you going to add something here?

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
MikeMcGarry: Ok

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: Okay.

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+            // Add the next response returned by api
+            AddMockHttpResponseToQueue(_httpListener, statusCode: expectedStatusCode, responseContent: "{\"some\": \"JsonResponseHere\"}");
+            var retryCount = 0;
+            RetryConfiguration.RetryPolicy =
+                Policy.Wrap(
+                    PollyApiRetryHandler.DefaultFallbackPolicy,
+                    Policy
+                        .HandleResult<IRestResponse>(PollyApiRetryHandler.GetInternalExceptionRetryCondition)
+                        .Retry(retryCount: 3, onRetry: (result, i) => retryCount++)
+                );
```
MikeMcGarry, line 51: As you are using both the `DefaultFallbackPolicy` and `GetInternalExceptionRetryCondition` from the `PollyApiRetryHandler` why not just use `RetryConfiguration.RetryPolicy = PollyApiRetryHandler.InternalExceptionRetryPolicyWithFallback;`?

This way you are testing that the `PollyApiRetryHandler` does not retry on a 400 response.

Or are you trying to isolate the `PollyApiRetryHandler.GetInternalExceptionRetryCondition`? If so you can just call this method with the inputs you want and don't even need to make an API call at all.

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+            Assert.That(retryCount, Is.EqualTo(0));
+        }
+
+        [Test]
+        public void CallGetPortfoliosApi_WhenHttpStatusIs200AndRetryConditionIsNotSatisfied_NoPollyRetryIsTriggered()
+        {
+            // It should do nothing when response code is 200
+            const int expectedStatusCode = 200;
+            // Add the next response returned by api
+            AddMockHttpResponseToQueue(_httpListener, expectedStatusCode, responseContent: "{\"some\": \"JsonResponseHere\"}");
```
MikeMcGarry, line 68: Why don't you just do `new Portfolio(...).ToJson()`?

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
MikeMcGarry: Note this pattern seems to be across multiple tests.

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+                    .HandleResult<IRestResponse>(response => response.StatusCode == (HttpStatusCode) returnedStatusCode)
+                    .Retry(expectedNumberOfRetries, onRetry: (result, i) => retryCount++));
+
+            // Calling GetPortfolio or any other API triggers the flow that triggers polly
+            var exception = Assert.Throws<ApiException>(
+                () => _apiFactory.Api<IPortfoliosApi>().GetPortfolio("any", "any")
+            );
+
+            Assert.That(retryCount, Is.EqualTo(expectedNumberOfRetries));
+            // Todo: In the future 0 error codes with throw an error after retries exceeded
```
MikeMcGarry, line 109: What does this mean in the context of this test?

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+
+            // TODO: This will still fail at deserialization, which currently is not handled and returns null.
+            // Will be done under a different PR. Response string will also be changed.
+            var sdkResponse = _apiFactory.Api<IPortfoliosApi>().GetPortfolio("any", "any");
+
+            Assert.That(retryCount, Is.EqualTo(0));
+        }
+
+        [Test]
+        public void CallGetPortfoliosApi_WhenApiResponseStatusCodeSatisfiesRetryCriteria_ExceedsPollyRetriesAndThrows()
```
MikeMcGarry, line 87: What is this test checking? Is it asserting that the `PollyApiRetryHandler.DefaultFallbackPolicy` works as expected? Is it asserting that you can write a Polly retry policy that retries on a certain status code? If so what does that demonstrate? You seem to be testing the functionality of Polly here rather than your own RetryHandler that you've created.

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+            var sdkResponse = _apiFactory.Api<IPortfoliosApi>().GetPortfolio("any", "any");
+
+            Assert.That(retryCount, Is.EqualTo(expectedNumberOfRetries));
+            // Todo: In the future 0 error codes with throw an error after retries exceeded
+            Assert.That(sdkResponse, Is.Null);
+        }
+
+        // Example of how an exponential backoff can be used with Polly
+        [Test]
+        public void CallGetPortfoliosApi_WhenApiResponseStatusCodeSatisfiesRetryCriteria_PollyRetryWithBackoffIsTriggered()
```
MikeMcGarry, line 166: Is this more of a tutorial than a test?

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+
+
+        [Test]
+        public void CreateLusidFactory_WhenRetryPolicyIsNull_AssignsDefaultRetryPolicy()
+        {
+            RetryConfiguration.RetryPolicy = null;
+
+            var newFactory = new LusidApiFactory(new Configuration());
+
+            Assert.That(RetryConfiguration.RetryPolicy, Is.Not.Null);
```
MikeMcGarry, line 351: Can we assert what it should be? Can we do this for sync and async?

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+        [Test]
+        public void CreateLusidFactory_WhenRetryPolicyIsAlreadyAssigned_ExistingRetryPolicyIsUsed()
+        {
+            var testPolicy = Policy.HandleResult<IRestResponse>(response => true).Retry();
+
+            RetryConfiguration.RetryPolicy = testPolicy;
+            var newFactory = new LusidApiFactory(new Configuration());
+
+            Assert.That(RetryConfiguration.RetryPolicy, Is.EqualTo(testPolicy));
+        }
```
MikeMcGarry, line 363: Nice

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
MikeMcGarry: Okay great, nice

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: Ok

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: That's okay let's stick with global. Looks to be difficult to modify this without modifying the ApiClient itself. You can do `Assert.That(RetryConfiguration.RetryPolicy, Is.EqualTo(PollyApiRetryHandler.InternalExceptionRetryPolicyWithFallback));`

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: Yes I meant the `LusidApiFactory`. 

Sounds good regarding the ExceptionFactoryHandler.

I think it is important to test that the `PollyApiRetryHandler` works as expected by actually using it and then asserting the results are as they should be. I can see we have some in-line Polly policies that is testing the Polly library more than it would be if we were to test our `PollyApiRetryHandler` class. 

It makes sense to have some examples on how a client can implement their own Polly handler, however if we are looking at this from the perspective of test coverage what we want to do is ensure that our `PollyApiRetryHandler` does what we expect. It isn't much value to set up an in-line Polly policy that retries on `502` as what is that showing? We want to use our retry handler and demonstrate how it handles a `502` error as every user of the SDK is going to be using the `PollyApiRetryHandler` by default going forward if they use the `LusidApiFactory`.

sdk/Lusid.Sdk/Utilities/ApiRetryHandler.cs
MikeMcGarry: Fair enough, let's handle this separately.

sdk/Lusid.Sdk/Utilities/PollyApiRetryHandler.cs
slacki123: Wasn't planning. This is more of an example for others who want to use polly

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: I am only isolating PollyApiRetryHandler.GetInternalExceptionRetryCondition as this is the only testable part.

Using the policy directly from PollyApiRetryHandler I cannot test that the number of retries was 0 as far as I'm aware. Maybe there's a way? wondering if there's a way to create a mock from a working object

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: That's a good shout actually!

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: Means that the response won't just be null, it will actually throw whenever we implement DefaultExceptionFactory

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: This is something absolutely outrageous that I've encountered. If there is no fallback policy, then after retries have exceeded, then `policyResult.Outcome` becomes  `OutcomeType.Failure`, here: 
https://github.com/finbourne/lusid-sdk-csharp-preview/blob/master/sdk/Lusid.Sdk/Client/ApiClient.cs#L494
which as a result constructs a response with absolutely NO INFORMATION to be used and to display to the user.

What I'm testing is that whenever we perhaps change any any default templates, upgrade the SDK or anything, the behaviour is as expected, that is, if retries have been exceeded, we then throw the error that would also be thrown if there were no retries.

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: Yes, since this is a client facing code and the proper way of implementing polly is tricky due to OpenAPI code, I think it's good to have these examples

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: cheers!

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: While it looks like it is testing Polly functionality, it's also testing that the behaviour remains as expected whenever we make any changes to default templates or upgrade the SDK.

Also because the way Polly works relies on OpenAPI generated code, it might not behave as someone would expect. One example is how lack of fallback policy gives you a null result rather than to rethrow the error it's retrying on. So these tests give all the necessary examples showing how Polly should properly be implemented

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
MikeMcGarry: If you want to test the `PollyApiRetryHandler` in its entirety you could pass a counter into `GetHttpResponseHandler` and have it incremented on each invocation.

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
MikeMcGarry: Ok

sdk/Lusid.Sdk/Utilities/PollyApiRetryHandler.cs
MikeMcGarry: Ok

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: Okay I understand, that's great. We should make sure that it is not at the expense of testing the `PollyApiRetryHandler` in its entirety. If you want to count invocations create a variable for counting retries and increment it in `GetHttpResponseHandler` by removing `static` so it can access the variable.

slacki123: genius

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: Done

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: Done

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
slacki123: Done

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
slacki123: Done x2

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: Resolved, nice!

sdk/Lusid.Sdk.Tests/Utilities/ApiRetryHandlerTests.cs
MikeMcGarry: Awesome, nice work!

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
MikeMcGarry: Sweet

sdk/Lusid.Sdk.Tests/Utilities/PollyApiRetryHandlerTests.cs
MikeMcGarry: Nice!

MikeMcGarry (approved):

PR #252, opened 2022-01-11
CTECH-672: Addition of Test_Transaction_Property (feature F13-4)

jaggerglyn:
# Pull Request Checklist

- [ ] Read the [contributing guidelines](../blob/master/docs/CONTRIBUTING.md)
- [ ] Tests pass

# Description of the PR

Addition of Test_Transaction_Property to check we can create a property on a transaction and successfully save the property information with the transaction.


sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
```diff
@@ -0,0 +1,117 @@
...
+        [SetUp]
+        public void SetUp()
+        {
+            var portfolioRequest = TestDataUtilities.BuildTransactionPortfolioRequest();
+            var portfolio = _transactionPortfoliosApi.CreatePortfolio(TestDataUtilities.TutorialScope, portfolioRequest);
+            _portfolioCode = portfolioRequest.Code;
+            Assert.That(portfolio?.Id.Code, Is.EqualTo(_portfolioCode));
+        }
+
+        [LusidFeature("F13-4")]
```
rizwansaeed, line 35: we should move this to https://github.com/finbourne/lusid-sdk-csharp-preview/blob/master/sdk/Lusid.Sdk.Tests/tutorials/Ibor/Transactions.cs alongside the other transaction examples

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
```diff
@@ -0,0 +1,117 @@
...
+        [Test]
+        public void Test_Transaction_Property()
```
rizwansaeed, line 37: these should be considered examples rather than tests so I would rename to be more descriptive, something like `Add_Transactions_To_Portfolio_With_Property`

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            var effectiveDate = new DateTimeOffset(2018, 1, 1, 0, 0, 0, TimeSpan.Zero);
+
+            //create the transaction request
+            string code = "TransactionTestDetail";
```
rizwansaeed, line 42: its helpful to have meaningful example data, so rather the dummy values, use something like ExecutingTrader = Glyn

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            try
+            {
+                propertyApi.GetPropertyDefinition("Transaction", TestDataUtilities.TutorialScope, code);
+            }
+            catch (ApiException)
+            {
+                //    Property definition doesn't exist (returns 404), so create one
+                //    Details of the property to be created
+                var propertyDefinition = new CreatePropertyDefinitionRequest(
```
rizwansaeed, line 102: i would check for the 404 before creating

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
```diff
@@ -0,0 +1,117 @@
...
+            string perpetualPropertyKey = $"Transaction/{TestDataUtilities.TutorialScope}/{code}";
+            string perpetualPropertyPropVal = "TestPropertyLabelValue";
+
+            EnsurePropertyDefinition(code);
+
+            var transaction = new TransactionRequest(
+                // unique transaction id
+                transactionId: Guid.NewGuid().ToString(),
+
+                // instruments must already exist in LUSID and have a valid LUSID instrument id
```
rizwansaeed, line 52: this isn't actually the case, we do allow for adding transactions without the instruments existing (see https://support.lusid.com/knowledgebase/article/KA-01860/en-us)

for now remove the comment

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
jaggerglyn: OK, this was originally placed in its own class and namespace to mirror Python. Now back with Transactions.cs

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
jaggerglyn: Done. Originally named to match Python.

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
jaggerglyn: Agreed.

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
jaggerglyn: Done. This code was originally copied from Instruments.cs and modified slightly so that the domain was Transaction rather than Instrument. 

I've just checked in a version that explicitly checks for errorcode == 404. I now also pass the domain as a parameter as I though we might extract this to TestDataUtilities.cs thereby also allowing the code to be removed from Instruments.cs. I haven't done this yet as I didn't know how you feel about having tutorial code in the Lusid.Sdk.Tests.Utilities namespace.

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
jaggerglyn: Comment removed.

sdk/Lusid.Sdk.Tests/tutorials/Properties/TransactionProperties.cs
rizwansaeed: In this case i think its fine. The purpose of the examples (aka tests) is to show a user how to craft a call to LUSID so the focus should be on clarify of code rather than the most efficient or being overly concise

PR #261, opened 2022-01-21
CTECH-189 hard code cut label names for deletion upon setup. refactor…

matthew-boardman:
… ClearCutLabel method so it can take any dictionary and doesn't throw if a cut label is not found

# Pull Request Checklist

- [x] Tests pass

# Description of the PR


rizwansaeed:
think we need a slightly different approach 
* remove the `ClearCutLabels` call in the `SetUp`
* wrap the `DeleteCutLabelDefinition` and `CreateCutLabelDefinition` in a `try-catch` so account for any concurrency issues
* keep the `ClearCutLabels` in the `TearDown`

rizwansaeed (approved):
merge once the build passes

PR #268, opened 2022-01-26
PLAT-784: SDK should throw exceptions on network connectivity issues

chrisbrook83:
# Pull Request Checklist

- [ ] Read the [contributing guidelines](../blob/master/docs/CONTRIBUTING.md)
- [ ] Tests pass

# Description of the PR
Under some circumstances we see SDK method calls returning null response values, which inevitably causes expected NullReferenceExceptions in calling code.

This PR makes the SDK throw an exception if the request fails to call the API at all (e.g. due to a network connectivity problem: bad URL / port etc), whereas it would previously return a null response.

### DIFF
--- a/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
+++ b/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
@@ -439,5 +439,12 @@ namespace Lusid.Sdk.Tests
             var date = apiResponse.GetRequestDateTime();
             Assert.IsNull(date);
         }
+
+        [Test]
+        public void ExceptionFactoryIsOverriddenWithCustomImplementation()
+        {
+            var api = _factory.Api<ApplicationMetadataApi>();
+            Assert.That(api.ExceptionFactory.GetInvocationList().Single().Method.Name, Is.EqualTo(nameof(LusidExceptionHandler.CustomExceptionFactory)));
+        }
     }
 }
--- a/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
+++ b/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
@@ -107,7 +107,29 @@ namespace Lusid.Sdk.Tests
                 () => new LusidApiFactory(apiConfig),
                 Throws.InstanceOf<UriFormatException>().With.Message.EqualTo("Invalid LUSID Uri: xyz"));
         }
-        
+
+        [Test]
+        public void NetworkConnectivityErrors_ThrowsException()
+        {
+            var apiConfig = ApiConfigurationBuilder.Build("secrets.json");
+            apiConfig.ApiUrl = "https://localhost:56789/api"; // nothing should be listening on this, so we should get a "No connection could be made" error...
+
+            var factory = new LusidApiFactory(apiConfig);
+            var api = factory.Api<PortfoliosApi>();
+            
+            Assert.That(
+                () => api.GetPortfolioWithHttpInfo("someScope", "someCode"),
+                Throws.InstanceOf<ApiException>()
+                    .With.Message.Contains("Error calling GetPortfolio: No connection could be made"));
+
+            // Note: these non-"WithHttpInfo" methods just unwrap the `Data` property from the call above.
+            // But these were the problematic ones, as they would previously just return a null value in this scenario.
+            Assert.That(
+                () => api.GetPortfolio("someScope", "someCode"),
+                Throws.InstanceOf<ApiException>()
+                    .With.Message.Contains("Error calling GetPortfolio")); // can't be more specific: get different exceptions locally vs in the build pipeline
+        }
+
         [Test]
         public void ApiResponse_CanExtract_RequestId()
         {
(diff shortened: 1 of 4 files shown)


sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
```diff
@@ -107,7 +107,29 @@ public void InvalidApiUrl_ThrowsException()
                 () => new LusidApiFactory(apiConfig),
                 Throws.InstanceOf<UriFormatException>().With.Message.EqualTo("Invalid LUSID Uri: xyz"));
         }
-        
+
+        [Test]
+        public void NetworkConnectivityErrors_ThrowsException()
```
slacki123, line 112: Great test, and looks good overall

Do we want to add some extra tests checking what happens on regular errors >400 errors  just for sanity?

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+            Exception defaultException = Configuration.DefaultExceptionFactory.Invoke(methodName, response);
+            if (defaultException != null)
+            {
+                return defaultException;
+            }
+
+            // Throw if ErrorText has been populated:
+            //  - Internal SDK deserialization errors will result in ErrorText to be not null.
+            //  - Network-level errors will also result in ErrorText being populated.
+            if (response.ErrorText != null)
```
slacki123, line 29: This is great and probably sufficient for now. We have to also keep in mind that with the current open API SDK version, there's a chance of an API deserialisation error

This leads to response.ErrorText being null, but the response.RawContent will be just an empty string

Maybe worth adding a comment for awareness?

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
```diff
@@ -76,6 +76,9 @@ public LusidApiFactory(Configuration configuration)
                     throw new Exception($"Unable to create type {api}");
                 }
 
+                // Replace the default implementation of the ExceptionFactory with a custom one defined by FINBOURNE
+                impl.ExceptionFactory = LusidExceptionHandler.CustomExceptionFactory;
```
slacki123, line 80: Do we want to test that APIs that are called indeed have their default exception factory overridden?

sdk/Lusid.Sdk/Utilities/LusidExceptionHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                return new ApiException(
+                    (int) response.StatusCode,
+                    $"Internal SDK error occurred when calling {methodName}: {response.ErrorText}",
+                    response.ErrorText,
+                    response.Headers
+                );
+            }
+
+            return null;
```
chrisbrook83, line 39: Note, I haven't ported across the behaviour here:
https://github.com/finbourne/lusid-sdk-csharp-preview/pull/230/files#diff-d62b6c2b0eb55f9d07ccad692c877fd73ff302448dfab7697134a801f5cfe685R41

...which is the handling for null response body values (due to deserialisation issues). 

Are there any calls which legitimately do not return a response body (only an http status code) - e.g. PUTs / POSTs?

slacki123: It is an observed behaviour where the JsonConvert doesn't actually fail when we pass an empty string or whitespace here
https://github.com/finbourne/lusid-sdk-csharp-preview/blob/master/sdk/Lusid.Sdk/Client/ApiClient.cs#L146

It in turn returns the same empty string or whitespace. 

Considering this is rather strange, and should maybe be investigated deeper, I am happy for this to be left out for now. 
The comment for awareness on the `ErrorText` check should suffice

slacki123: That's because by default JsonConvert doesn't have `PropertyRequired` set by default

https://www.newtonsoft.com/json/help/html/JsonPropertyRequired.htm
https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Required.htm

sdk/Lusid.Sdk/Utilities/LusidApiFactory.cs
chrisbrook83: Good idea. Have added `ExceptionFactoryIsOverriddenWithCustomImplementation`

sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
chrisbrook83: I've now pulled over the LusidExceptionHandlerTests, so we've got tests like `CallCustomExceptionFactory_FailsWithApiError_ReturnsTheSameErrorCodeAsApi` (testing 404, for example)

sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
```diff
@@ -414,5 +439,12 @@ public void ApiResponseInvalidDateHeader_ReturnsNull_DateHeader()
             var date = apiResponse.GetRequestDateTime();
             Assert.IsNull(date);
         }
+
+        [Test]
+        public void ExceptionFactoryIsOverriddenWithCustomImplementation()
+        {
+            var api = _factory.Api<ApplicationMetadataApi>();
+            Assert.That(api.ExceptionFactory.GetInvocationList().Single().Method.Name, Is.EqualTo(nameof(LusidExceptionHandler.CustomExceptionFactory)));
```
slacki123, line 447: very nice!

PR #290, opened 2022-04-13
ANA-910: Add example for InterestRateSwap lifecycle handling

SamuelColvinFinbourne:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

### DIFF
--- a/sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
@@ -118,6 +118,7 @@ namespace Lusid.Sdk.Tests.Tutorials.Instruments
             CallLusidGetPortfolioCashFlowsEndpoint(irs, model);
         }
 
+        [LusidFeature("F22-46")]
         [TestCase(ModelSelection.ModelEnum.ConstantTimeValueOfMoney)]
         [TestCase(ModelSelection.ModelEnum.Discounting)]
         public void LifeCycleManagementForInterestRateSwap(ModelSelection.ModelEnum model)
--- a/sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
@@ -20,7 +20,7 @@ namespace Lusid.Sdk.Tests.Tutorials.Instruments
             // For LUSID to pick up these quotes, we have added a RIC rule to the recipe (see BuildRecipeRequest in TestDataUtilities.cs) 
             // The RIC rule has a large quote interval, this means that we can use one reset quote for all the resets.
             // For accurate pricing, one would want to upsert a quote per reset. 
-            var quoteRequest = TestDataUtilities.BuildQuoteRequest("USD6M", QuoteSeriesId.InstrumentIdTypeEnum.ClientInternal, 0.05m, "USD", TestDataUtilities.EffectiveAt);
+            var quoteRequest = TestDataUtilities.BuildQuoteRequest("BP00", QuoteSeriesId.InstrumentIdTypeEnum.RIC, 0.05m, "USD", TestDataUtilities.EffectiveAt);
             var upsertResponse = _quotesApi.UpsertQuotes(scope, quoteRequest);
             Assert.That(upsertResponse.Failed.Count, Is.EqualTo(0));
             Assert.That(upsertResponse.Values.Count, Is.EqualTo(quoteRequest.Count));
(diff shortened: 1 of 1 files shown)


sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -123,5 +123,124 @@ public void InterestRateSwapPortfolioCashFlowsExample(ModelSelection.ModelEnum m
...
+
+            // We expect that the PV of the InterestRateSwap should be zero after the maturity.
+            // CREATE valuation request for this portfolio consisting of the InterestRateSwap,
+            // with valuation dates a few days before, day of and a few days after the instrument expiration.
+            var valuationRequest = TestDataUtilities.CreateValuationRequest(
+                scope,
+                portfolioCode,
+                recipeCode,
+                effectiveAt: interestRateSwap.MaturityDate.AddDays(2).AddMilliseconds(1),
+                effectiveFrom: interestRateSwap.MaturityDate.AddDays(-2).AddMilliseconds(1)); // TODO ANA-910: should I mention the millisecond issue?
```
jz-finbourne, line 160: I think we should remove the comment and just the existence of the jira is sufficient.
Keeping it public doesn't look good and we may forget to remove when it is done.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -123,5 +123,124 @@ public void InterestRateSwapPortfolioCashFlowsExample(ModelSelection.ModelEnum m
...
+            var valuationAfterUpsertingCashFlows = _aggregationApi.GetValuation(valuationRequest);
+
+            // ASSERT that we have some cash in the portfolio
+            var luidsContainedInValuationResponse = valuationAfterUpsertingCashFlows
+                .Data
+                .Select(dictionaryOfMetrics => (string)dictionaryOfMetrics[TestDataUtilities.Luid]);
+            Assert.That(luidsContainedInValuationResponse, Does.Contain($"CCY_{currency}"));
+
+
+            // Everytime a cash flow goes transaction occurs, the PV changes accordingly in an equal and opposite amount.
```
jz-finbourne, line 214: "Everytime a cash flow goes transaction occurs"...?
--> PV of the IRS

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -123,5 +123,124 @@ public void InterestRateSwapPortfolioCashFlowsExample(ModelSelection.ModelEnum m
...
+            // Thus we expect the PV to be constant throughout.
+
+            // Note all the dates on which cash flows occur.
+            var cashFlowDates = allInterestRateSwapCashFlows
+                .Select(cashFlowTransaction => cashFlowTransaction.TransactionDate)
+                .Distinct()
+                .ToList();
+
+            foreach (var cashFlowDate in cashFlowDates)
```
jz-finbourne, line 223: optional: add a small comment above this foreach loop to explain

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -123,5 +123,124 @@ public void InterestRateSwapPortfolioCashFlowsExample(ModelSelection.ModelEnum m
...
+            var valuationBeforeAndAfterExpirationInterestRateSwap = _aggregationApi.GetValuation(valuationRequest);
+            TestDataUtilities.CheckNoCashPositionsInValuationResults(
+                valuationBeforeAndAfterExpirationInterestRateSwap,
+                currency);
+            TestDataUtilities.CheckNonZeroPvBeforeMaturityAndZeroAfter(
+                valuationBeforeAndAfterExpirationInterestRateSwap,
+                interestRateSwap.MaturityDate);
+
+            // GET all upsertable cashflows (transactions) for the InterestRateSwap.
+            // EffectiveAt after maturity so we have all the data.
```
jz-finbourne, line 174: suggest "data" -> "reset quotes" or alike

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -123,5 +123,124 @@ public void InterestRateSwapPortfolioCashFlowsExample(ModelSelection.ModelEnum m
             var irs = InstrumentExamples.CreateExampleInterestRateSwap("Vanilla");
             CallLusidGetPortfolioCashFlowsEndpoint(irs, model);
         }
+
+        [TestCase(ModelSelection.ModelEnum.ConstantTimeValueOfMoney)]
```
jz-finbourne, line 127: think you'll need to add a lusid feature number

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
SamuelColvinFinbourne: changed to "so we have all the cash flows"

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
SamuelColvinFinbourne: added "CHECK that the PV of the portfolio is constant across dates that are in a small neighbourhood of every cash flow transaction date"

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -123,5 +123,124 @@ public void InterestRateSwapPortfolioCashFlowsExample(ModelSelection.ModelEnum m
...
+            // with valuation dates a few days before, day of and a few days after the instrument expiration.
+            var valuationRequest = TestDataUtilities.CreateValuationRequest(
+                scope,
+                portfolioCode,
+                recipeCode,
+                effectiveAt: interestRateSwap.MaturityDate.AddDays(2).AddMilliseconds(1),
+                effectiveFrom: interestRateSwap.MaturityDate.AddDays(-2).AddMilliseconds(1)); // TODO ANA-910: should I mention the millisecond issue?
+
+            // CALL GetValuation before upserting back the cash flows. We check
+            // (1) there is no cash holdings in the portfolio prior to expiration
```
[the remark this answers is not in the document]
SamuelColvinFinbourne: Changed to "prior to upserting the cash flows"

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
SamuelColvinFinbourne: added

sdk/Lusid.Sdk.Tests/tutorials/Instruments/InterestRateSwapExamples.cs
```diff
@@ -20,7 +20,7 @@ protected override void CreateAndUpsertMarketDataToLusid(string scope, ModelSele
             // For LUSID to pick up these quotes, we have added a RIC rule to the recipe (see BuildRecipeRequest in TestDataUtilities.cs) 
             // The RIC rule has a large quote interval, this means that we can use one reset quote for all the resets.
             // For accurate pricing, one would want to upsert a quote per reset. 
-            var quoteRequest = TestDataUtilities.BuildQuoteRequest("USD6M", QuoteSeriesId.InstrumentIdTypeEnum.ClientInternal, 0.05m, "USD", TestDataUtilities.EffectiveAt);
+            var quoteRequest = TestDataUtilities.BuildQuoteRequest("BP00", QuoteSeriesId.InstrumentIdTypeEnum.RIC, 0.05m, "USD", TestDataUtilities.EffectiveAt);
```
SamuelColvinFinbourne, line 23: This change is so the built quote has identifiers that match what is needed to be a reset for the standard example of a swap. It is likely the reason this worked before, and didn't break any other tests by changing, is because no valuation requests on swaps required a reset. i.e. all our valuations were purely forward looking and didn't require any past data.

jz-finbourne: Am little surprised the existing test (before your changes) was purely forward looking and didn't require reset. I recall hacking at this... I wonder if it is because we've changed the underlying swap and/or effectiveAt since my merge.... I'm wondering if "Vanilla" swap refers to a different swap to the one I used....

rizwansaeed: what is the corresponding change in the recipe for changing the quote provider type? where does a user see this?

SamuelColvinFinbourne: @jz-finbourne This no longer uses Nick's refactor, so there is no concept of "Vanilla" or otherwise yet. At the moment there is just the one standard swap.
It looks like valuation calls are generically being made effective 2020-01-02, but the instrument's start date is 2020-02-07, so no resets required.

SamuelColvinFinbourne: @rizwansaeed There is no change needed in the recipe. This quote was not actually being used in the tests. 
Currently, we set up an extremely generic recipe using a `CreateAndUpsertRecipe` method. This already contains a rule that says we can look back up to 10 years for any RIC identified resets.
So when I value the instrument in the middle of its lifecycle, I see that I need a RIC reset for the index, the RIC rule then gets applied, sees this as the only quote in the last 10 years, then uses it.

rizwansaeed: ok, we can leave for now, this will relate to the issue of how a user sets up recipes, the current implication from all the current examples is that they need _everything_ for a call to work

rizwansaeed (approved):

PR #291, opened 2022-04-14
ANA-909: Lifecycle for FX options

tfnc-finbourne:
# Pull Request Checklist

- [x ] Read the [contributing guidelines](../blob/master/docs/CONTRIBUTING.md)
- [ x] Tests pass
- [ x ] Raised the PR against the `develop` branch

# Description of the PR

### DIFF
--- a/sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
+++ b/sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
@@ -52,7 +52,7 @@ namespace Lusid.Sdk.Tests.Utilities
         internal static LusidInstrument CreateExampleEquityOption(bool isCashSettled = false)
             => new EquityOption(
                 startDate: new DateTimeOffset(2020, 2, 7, 0, 0, 0, TimeSpan.Zero),
-                optionMaturityDate: new DateTimeOffset(2020, 12, 19, 0, 0, 0, TimeSpan.Zero),
+                optionMaturityDate: new DateTimeOffset(2020, 12, 19, 0, 0, 0, TimeSpan.Zero), // note that this is a weekend; lusid may skip over this date when generating valuation schedules
                 optionSettlementDate: new DateTimeOffset(2020, 12, 21, 0, 0, 0, TimeSpan.Zero),
                 deliveryType: isCashSettled
                     ? "Cash"
--- a/sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
+++ b/sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
@@ -19,6 +19,7 @@ namespace Lusid.Sdk.Tests.Utilities
         public static string ValuationPvKey = "Valuation/PV/Amount";
         public static string InstrumentName = "Instrument/default/Name";
         public static readonly string ValuationPv = "Valuation/PV/Amount";
+        public static readonly string ValuationPvInReportCcy = "Valuation/PvInReportCcy";
         public static readonly string Luid = "Instrument/default/LusidInstrumentId";
         public static readonly string Currency = "Valuation/PV/Ccy";
         
@@ -28,6 +29,7 @@ namespace Lusid.Sdk.Tests.Utilities
             new AggregateSpec(ValuationDateKey, AggregateSpec.OpEnum.Value),
             new AggregateSpec(InstrumentName, AggregateSpec.OpEnum.Value),
             new AggregateSpec(ValuationPvKey, AggregateSpec.OpEnum.Value),
+            new AggregateSpec(ValuationPvInReportCcy, AggregateSpec.OpEnum.Value),
             new AggregateSpec(InstrumentTag, AggregateSpec.OpEnum.Value),
             new AggregateSpec(Luid, AggregateSpec.OpEnum.Value),
             new AggregateSpec(Currency, AggregateSpec.OpEnum.Value)
@@ -220,7 +222,7 @@ namespace Lusid.Sdk.Tests.Utilities
         /// <summary>
         /// This method upserts JPY/USD and USD/JPY FX quotes for every day in the date range.
         /// </summary>
-        public static Dictionary<string, UpsertQuoteRequest> BuildFxRateRequest(DateTimeOffset effectiveFrom,  DateTimeOffset effectiveAt, bool useConstantFxRate = false)
+        public static Dictionary<string, UpsertQuoteRequest> BuildFxRateRequest(DateTimeOffset effectiveFrom, DateTimeOffset effectiveAt, bool useConstantFxRate = false)
         {
             // CREATE FX quotes and inverse fx rate in the desired date range
             var upsertQuoteRequests = new Dictionary<string, UpsertQuoteRequest>();
@@ -681,7 +683,7 @@ namespace Lusid.Sdk.Tests.Utilities
             var pvsAcrossDates = result
                 .Data // Access a list of result dictionaries
                 .GroupBy(d => (DateTime) d[ValuationDateKey]) // We group by date
-                .Select(pvGroup => pvGroup.Sum(record => (double) record[ValuationPv])); // we pick up the PV
+                .Select(pvGroup => pvGroup.Sum(record => (double) record[ValuationPvInReportCcy])); // we pick up the PV
 
             var isWithinTolerance = ValuesWithinARelativeDiffTolerance(pvsAcrossDates, relativeDifferenceTolerance);
             
@@ -719,9 +721,13 @@ namespace Lusid.Sdk.Tests.Utilities
                 {
                     Assert.That(pv, Is.Not.EqualTo(0).Within(1e-12));
                 }
-                else
+                if (date > maturityDate)
                 {
                     Assert.That(pv, Is.EqualTo(0).Within(1e-12));
+                }
+                if (date == maturityDate)
+                { // ANA-1292: Currently, we do not perform an assertion if the date is exactly the maturity date, as it's not clear what the PV should be.
+                    continue;
                 }
             }
         }
--- a/sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
+++ b/sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
@@ -140,7 +140,7 @@ namespace Lusid.Sdk.Tests.tutorials.Ibor
             Assert.That(amountsPositive, Is.True);
             
             // GIVEN the cashflow transactions, we create from them transaction requests and upsert them.
-            var upsertCashFlowTransactions = PopulateCashFlowTransactionWithUniqueIds(cashFlows.Values, bond.DomCcy);
+            var upsertCashFlowTransactions = PopulateCashFlowTransactionWithUniqueIds(cashFlows.Values);
             _transactionPortfoliosApi.UpsertTransactions(_portfolioScope, _portfolioCode, MapToCashFlowTransactionRequest(upsertCashFlowTransactions));
 
             var expectedPortfolioTransactions = _transactionPortfoliosApi.GetTransactions(
@@ -214,7 +214,7 @@ namespace Lusid.Sdk.Tests.tutorials.Ibor
             Assert.That(currencyAndAmounts, Is.EquivalentTo(expectedCashFlows)); 
             
             // GIVEN the cashflow transactions, we create from them transaction requests and upsert them.
-            var upsertCashFlowTransactions = PopulateCashFlowTransactionWithUniqueIds(cashFlows.Values, fxForward.DomCcy);
+            var upsertCashFlowTransactions = PopulateCashFlowTransactionWithUniqueIds(cashFlows.Values);
             _transactionPortfoliosApi.UpsertTransactions(_portfolioScope, _portfolioCode, MapToCashFlowTransactionRequest(upsertCashFlowTransactions));
 
             _recipeApi.DeleteConfigurationRecipe(_portfolioScope, modelRecipeCode);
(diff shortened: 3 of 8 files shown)


sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
```diff
@@ -19,6 +19,7 @@ public static class TestDataUtilities
         public static string ValuationPvKey = "Valuation/PV/Amount";
         public static string InstrumentName = "Instrument/default/Name";
         public static readonly string ValuationPv = "Valuation/PV/Amount";
+        public static readonly string ValuationPvInReportCcy = "Valuation/PvInReportCcy";
```
tfnc-finbourne, line 22: Necessary to get portfolio PV when there's multiple currencies of cash in the portfolio.

sdk/Lusid.Sdk.Tests/tutorials/Ibor/PortfolioCashFlows.cs
```diff
@@ -243,11 +243,11 @@ internal static List<TransactionRequest> MapToCashFlowTransactionRequest(IEnumer
         // Given a transaction, this method creates a TransactionRequest so that it can be upserted back into LUSID.
         // InstrumentUid is additionally added to identify where the cashflow came from. The transaction ID needs to
         // be unique.
-        internal static IEnumerable<Transaction> PopulateCashFlowTransactionWithUniqueIds(IEnumerable<Transaction> transactions, string cashFlowCurrency)
+        internal static IEnumerable<Transaction> PopulateCashFlowTransactionWithUniqueIds(IEnumerable<Transaction> transactions)
         {
             foreach (var transaction in transactions)
             {
-                transaction.InstrumentIdentifiers.Add("Instrument/default/Currency", cashFlowCurrency);
+                transaction.InstrumentIdentifiers.Add("Instrument/default/Currency", transaction.TransactionCurrency);
```
tfnc-finbourne, line 250: refactor, don't see why we need a separate parameter specifying cashflow ccy when the transaction should know this info

sdk/Lusid.Sdk.Tests/tutorials/Instruments/EquityOptionExamples.cs
```diff
@@ -257,7 +256,7 @@ public void LifeCycleManagementForCashSettledEquityOption(ModelSelection.ModelEn
             var containsCashAfterUpsertion = valuationAfterUpsertingCashFlows
                 .Data
                 .Select(d => (string) d[TestDataUtilities.Luid])
-                .Any(luid => luid != $"CCY_{equityOption.DomCcy}");
+                .Any(luid => luid == $"CCY_{equityOption.DomCcy}");
```
tfnc-finbourne, line 259: old code was bugged it seems

sdk/Lusid.Sdk.Tests/tutorials/Instruments/EquityOptionExamples.cs
```diff
@@ -238,12 +238,11 @@ public void LifeCycleManagementForCashSettledEquityOption(ModelSelection.ModelEn
                 equityOption.DomCcy);
             TestDataUtilities.CheckNonZeroPvBeforeMaturityAndZeroAfter(
                 valuationBeforeAndAfterExpirationEquityOption,
-                equityOption.OptionMaturityDate);
+                equityOption.OptionSettlementDate);
```
tfnc-finbourne, line 241: the pv actually drops off at settlement, not maturity, so changed here

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
```diff
@@ -719,10 +721,14 @@ internal static void CheckNonZeroPvBeforeMaturityAndZeroAfter(ListAggregationRes
                 {
                     Assert.That(pv, Is.Not.EqualTo(0).Within(1e-12));
                 }
-                else
+                if (date > maturityDate)
                 {
                     Assert.That(pv, Is.EqualTo(0).Within(1e-12));
                 }
+                if (date == maturityDate)
+                { // ANA-1292: Currently, we do not perform an assertion if the date is exactly the maturity date, as it's not clear what the PV should be.
```
jz-finbourne, line 729: I don't prefer adding ANA-XXXX here on GH as it is public.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
```diff
@@ -150,5 +152,110 @@ public void FxOptionPortfolioCashFlowsExample(ModelSelection.ModelEnum model, bo
             var fxOption = InstrumentExamples.CreateExampleFxOption(isDeliveryNotCash);
             CallLusidGetPortfolioCashFlowsEndpoint(fxOption, model);
         }
+
+        /// <summary>
+        /// Lifecycle management (i.e. conservation of PV on cashflow events) for fx option
```
jz-finbourne, line 157: For these public docs, can we do "FX" and "PV" over its lowercase counterpart please? 
(I'm sure I've been guilty of otherwise)

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
```diff
@@ -150,5 +152,110 @@ public void FxOptionPortfolioCashFlowsExample(ModelSelection.ModelEnum model, bo
...
+        /// Cash-settled case:
+        ///      pv(option) = max(spot - strike, 0) before maturity and
```
jz-finbourne, line 159: Believe `pv(option) = max(spot - strike, 0) ` is only true for CTVoM
Did I write analogous comment for EQ option?

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
```diff
@@ -150,5 +152,110 @@ public void FxOptionPortfolioCashFlowsExample(ModelSelection.ModelEnum model, bo
...
+        ///      pv(cashflow) = max(spot_at_maturity - strike, 0) after maturity
+        /// Physically-settled case: If option is exercised, pay the strike to obtain the underlying.
+        ///      pv(option) = max(spot - strike, 0) before maturity and
+        ///      pv(option exercise payment) + pv(received underlying) = max(spot_at_maturity - strike, 0) after maturity (assuming fx rates haven't changed)
```
jz-finbourne, line 163: not actually too sure on the wording of pv(option exercise payment) and pv(received underlying).

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
jz-finbourne: Ah right, reading test, it's only for CTVoM right now..hmm

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
```diff
@@ -150,5 +152,110 @@ public void FxOptionPortfolioCashFlowsExample(ModelSelection.ModelEnum model, bo
...
+                .Data
+                .Select(d => (string)d[TestDataUtilities.Luid])
+                .Where(luid => luid.StartsWith("CCY_"))
+                .Distinct();
+            Assert.That(currenciesInPortfolioAfterUpsertion.Count, Is.EqualTo(isDeliveryNotCash ? 2 : 1));
+
+            // ASSERT portfolio PV is constant for each valuation date.
+            // We expect this to be true since we upserted the cashflows back in.
+            // That is instrument pv + cashflow = option payoff = constant for each valuation date.
+            TestDataUtilities.CheckPvIsConstantAcrossDatesWithinTolerance(valuationAfterUpsertingCashFlows, relativeDifferenceTolerance: 1e-10);
```
jz-finbourne, line 253: may worth adding (again) comment that this is because fx rate is const.

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
```diff
@@ -150,5 +152,110 @@ public void FxOptionPortfolioCashFlowsExample(ModelSelection.ModelEnum model, bo
...
+            var (instrumentID, portfolioCode) = CreatePortfolioAndInstrument(scope, option);
+
+            // UPSERT option to portfolio and populating stores with required market data.
+            CreateAndUpsertMarketDataToLusid(scope, model, option);
+
+            // CREATE recipe to price the portfolio with
+            var recipeCode = CreateAndUpsertRecipe(scope, model, windowValuationOnInstrumentStartEnd: true);
+
+            // GET all upsertable cashflows (transactions) for the option
+            // For this test, we will be performing valuations around the option settlement date, so we upsert some spot rates for the underlying fx rates.
```
jz-finbourne, line 188: "FX" here also

tfnc-finbourne: done

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
tfnc-finbourne: done

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
tfnc-finbourne: removed

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
tfnc-finbourne: just changed to pv(cash balance)

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
tfnc-finbourne: done

sdk/Lusid.Sdk.Tests/tutorials/Instruments/EquityOptionExamples.cs
tfnc-finbourne: good point, am ignoring tho since that’s just correcting a different test. would change it if it was my test

sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
```diff
@@ -52,7 +52,7 @@ internal static LusidInstrument CreateExampleFxOption(bool isDeliveryNotCash = t
         internal static LusidInstrument CreateExampleEquityOption(bool isCashSettled = false)
             => new EquityOption(
                 startDate: new DateTimeOffset(2020, 2, 7, 0, 0, 0, TimeSpan.Zero),
-                optionMaturityDate: new DateTimeOffset(2020, 12, 19, 0, 0, 0, TimeSpan.Zero),
+                optionMaturityDate: new DateTimeOffset(2020, 12, 19, 0, 0, 0, TimeSpan.Zero), // note that this is a weekend; lusid may skip over this date when generating valuation schedules
```
rizwansaeed, line 55: for a user we need to define what the expected behaviour is, "may" is too ambiguous

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
```diff
@@ -719,10 +721,14 @@ internal static void CheckNonZeroPvBeforeMaturityAndZeroAfter(ListAggregationRes
                 {
                     Assert.That(pv, Is.Not.EqualTo(0).Within(1e-12));
                 }
-                else
+                if (date > maturityDate)
                 {
                     Assert.That(pv, Is.EqualTo(0).Within(1e-12));
                 }
+                if (date == maturityDate)
+                { // Currently, we do not perform an assertion if the date is exactly the maturity date, as it's not clear what the PV should be.
```
rizwansaeed, line 729: _"as it's not clear what the PV should be"_ doesnt help a user, we need to make a statement on what the user should expect

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
```diff
@@ -150,5 +152,111 @@ public void FxOptionPortfolioCashFlowsExample(ModelSelection.ModelEnum model, bo
...
+
+            // CREATE portfolio and add instrument to the portfolio
+            var scope = Guid.NewGuid().ToString();
+            var (instrumentID, portfolioCode) = CreatePortfolioAndInstrument(scope, option);
+
+            // UPSERT option to portfolio and populating stores with required market data.
+            CreateAndUpsertMarketDataToLusid(scope, model, option);
+
+            // CREATE recipe to price the portfolio with
+            var recipeCode = CreateAndUpsertRecipe(scope, model, windowValuationOnInstrumentStartEnd: true);
```
rizwansaeed, line 186: as a general comment this approach is very confusing to a new user, the implication is that you need _all_ the items in the recipe generated by`TestDataUtilities.BuildRecipeRequest`, or put another way, what is the minimum requirement for a recipe to generate the cashflows?

rizwansaeed (changes requested):
inline comments

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
tfnc-finbourne: I would agree with you but this is the way it's currently set up, unless you want to completely rethink how we do examples?

rizwansaeed: I think we need a specific (minimal) recipe per instrument/use case, for now maybe just do it for this PR and raise a jira to retrofit. Otherwise you may end up field lots of support calls from clients

sdk/Lusid.Sdk.Tests/Utilities/InstrumentExamples.cs
tfnc-finbourne: clarified a bit

sdk/Lusid.Sdk.Tests/Utilities/TestDataUtilities.cs
tfnc-finbourne: to some degree it's a comment for us as much as it is for the user, but reworded to clarify that it's a decision we've not yet made

sdk/Lusid.Sdk.Tests/tutorials/Instruments/FxOptionExamples.cs
tfnc-finbourne: Okay, for the record Market Data also works like this, CreateAndUpsertMarketDataToLusid also sometimes upserts more than the minimal set necessary for the use case.

Let me discuss with team about this and get back to you then

tfnc-finbourne: @rizwansaeed 
After discussion it seems that we in general agree with you, however the priority for now is to get these examples and instrument coverage in. We'll revisit these later, possibly after a cleanup of recipes and training up the rest of the team on the specifics of how they work.

PR #307, opened 2022-06-14
ANA-566: Fix wrong tests

SamuelColvinFinbourne:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR


SamuelColvinFinbourne:
These tests were relying on an existing bug that has been fixed. Essentially the portfolio was being build at the end of the window of cashflows being asked for instead of at the effectiveAt. This means that you could ask for cashflows of the portfolio effective before any transactions had been added and receive the cashflows for instruments not yet in the portfolio. The fix is to move the effectiveAt to be the same as (or later than) the date that the transaction occurs on the portfolio.

fbn-seamus:
Looks good to me, I don't think I have approval rights here or know if they're needed. @slemasne could you approve if needed?

PR #326, opened 2022-08-02
Feature/ctech 1502.add timeout option via api factory

fox8090:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Adds optional parameter to set the `timeout` property on the `configuration`.

### DIFF
--- a/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
+++ b/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
@@ -523,5 +523,32 @@ namespace Lusid.Sdk.Tests
             // assert
             Assert.IsEmpty(api.Configuration.DefaultHeaders);
         }
+
+        [Test]
+        public void ApiFactoryWithTimeout() 
+        {
+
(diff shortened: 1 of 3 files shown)


sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs
```diff
@@ -523,5 +523,32 @@ public void ApiFactoryHasNoDefaultHeadersWhenNoneAreSpecified()
...
+
+            var factory = LusidApiFactoryBuilder.Build("secrets.json", timeout);
+
+            var api = factory.Api<ScopesApi>();
+            DateTime start = DateTime.Now;
+
+            Assert.That(() => api.ListScopes(null), Throws.InstanceOf<ApiException>().With.Message.EqualTo("Internal SDK error occurred when calling ListScopes: The operation has timed out."));
+
+            try
+            {
```
jyaffe, line 542: do you need to call the api again?

is it sufficient to just record the end time after the first assert and then run your second assert?

PR #337, opened 2022-11-11
CTECH-2254 Update the timeout to 2 minutes to deal with some prevalen…

JeremyBarber:
…t timeouts under normal operating conditions

# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

We are getting frequent client reports of timeouts against the calendars API under what we would consider to be very normal loads. Specifically, the timeouts are being generated from the C# SDK, not LUSID itself.

LUSID is responding with success codes roughly ten seconds after the SDK has timed out. So, in order to resolve this issue, we are raising the timeout threshold in the SDK from 100 seconds to 120 seconds.

The new timeout threshold should give us more consistent behaviour with LUSID and alleviate the issues observed by clients.

### DIFF
--- a/sdk/Lusid.Sdk/Client/Configuration.cs
+++ b/sdk/Lusid.Sdk/Client/Configuration.cs
@@ -118,7 +118,7 @@ namespace Lusid.Sdk.Client
             };
 
             // Setting Timeout has side effects (forces ApiClient creation).
-            Timeout = 100000;
+            Timeout = 120000;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


sdk/Lusid.Sdk/Client/Configuration.cs
```diff
@@ -118,7 +118,7 @@ public Configuration()
             };
 
             // Setting Timeout has side effects (forces ApiClient creation).
-            Timeout = 100000;
+            Timeout = 120000;
```
rizwansaeed, line 121: this is part of the generated code and will get overridden each time the sdk is generated

rizwansaeed:
i think it might be better to update the client code to set the timeout as needed, something like https://github.com/finbourne/lusid-sdk-csharp-preview/blob/master/sdk/Lusid.Sdk.Tests/LusidApiFactoryTests.cs#L533

JeremyBarber:
Had a good discusion about the intent behind the timeouts and I agree with the idea that large calendar operations should be given their own timeout in their API config

